PR #1, opened 2018-06-14
Initial project structure

fuzzball81:
Initial project structure with travis/pipenv/tox support.


CONTRIBUTING.rst
```diff
@@ -0,0 +1,128 @@
...
+------------
+
+Ready to contribute? Here's how to set up `git_wrapper` for local development.
+
+1. Fork the `git_wrapper` repo on GitHub.
+2. Clone your fork locally::
+
+    $ git clone git@github.com:your_name_here/git_wrapper.git
+
+3. Install your local copy into a virtualenv. Assuming you have virtualenvwrapper installed, this is how you set up your fork for local development::
```
jguiditta: If you are using pipenv, you may want to tweak these setup directions, though that could be in a subsequent patch

CONTRIBUTING.rst
```diff
@@ -64,11 +64,10 @@ Ready to contribute? Here's how to set up `git_wrapper` for local development.
 
     $ git clone git@github.com:your_name_here/git_wrapper.git
 
-3. Install your local copy into a virtualenv. Assuming you have virtualenvwrapper installed, this is how you set up your fork for local development::
+3. Use pipenv to setup a new virtualenv. Pipenv will create a new virtualenv with packages at the correct versions. Assuming you have pipenv installed, this is how you set up your fork for local development::
 
-    $ mkvirtualenv git_wrapper
-    $ cd git_wrapper/
-    $ python setup.py develop
+    $ pipenv install --dev --ignore-pipfile
```
jguiditta: Two questions here:
1. why ignore the pipfile?
2. why not use 'make dev' like we do in koji_wrapper?

CONTRIBUTING.rst
```diff
@@ -80,10 +79,9 @@ Ready to contribute? Here's how to set up `git_wrapper` for local development.
    tests, including testing other Python versions with tox::
 
     $ flake8 git_wrapper tests
-    $ python setup.py test or py.test
+    $ py.test tests
```
jguiditta: If you want to call pytest directly, you can do so simply with 'pytest' and it will find the standard directories.  Alternatively, you can have people run 'make test' for their current pipenv, or 'make test-all' to run tox against all supported versions

CONTRIBUTING.rst
```diff
@@ -78,8 +79,8 @@ Ready to contribute? Here's how to set up `git_wrapper` for local development.
 5. When you're done making changes, check that your changes pass flake8 and the
    tests, including testing other Python versions with tox::
 
-    $ flake8 git_wrapper tests
-    $ py.test tests
+    $ make lint
+    $ make tests
```
jguiditta: Looks good, with minor change that the make target is 'test' rather than 'tests'

PR #4, opened 2018-06-20
Adding GitWrapperCherry

fuzzball81:
- Added a base class and tests that include basic git functionality.
- Added a flake8 tox environment and updated travis to support it
- Removed travis deployment for now.
- Removed git_wrapper.py and test_git_wrapper.py in favor of base.

### DIFF
--- a/git_wrapper/cherry.py
+++ b/git_wrapper/cherry.py
@@ -1,0 +1,34 @@
+#! /usr/bin/env python
+'''This module acts as an interface for common git tasks'''
+
+import re
+
+from git_wrapper.base import GitWrapperBase
+
+
+class GitWrapperCherry(GitWrapperBase):
+    '''Provides git cherry functionality'''
+
+    def __init__(self, path='', repo=None):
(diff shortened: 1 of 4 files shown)


git_wrapper/cherry.py
```diff
@@ -0,0 +1,34 @@
...
+
+class GitWrapperCherry(GitWrapperBase):
+    '''Provides git cherry functionality'''
+
+    def __init__(self, path='', repo=None):
+        '''Constructor for GitWrapperCherry'''
+        super(GitWrapperCherry, self).__init__(path=path, repo=repo)
+        self.__new_regex = re.compile(r'^\+\s(.*?)\s(.*)')
+        self.__both_regex = re.compile(r'^\-\s(.*?)\s(.*)')
+
```
jpichon: This isn't a review comment, only a question: most of the codebases I worked with were happy with the single underscore convention for private attributes, because the double underscore name mangling was too much trouble I think. I'm wondering about the conventions in this library? Is it used when indicating class-type attributes usually?

PR #6, opened 2018-06-27
'Make lint' should match the tox setup

jpichon:
Currently, 'make lint' fails because it's lacking the same 'ignore' rules set up in the tox flake8 job.


fuzzball81:
Good catch!

PR #12, opened 2018-07-03
Fixing mergify config file whitespace

fuzzball81:


jpichon:
I think maybe it's an issue with Mergify (opened https://github.com/Mergifyio/mergify-engine/issues/65 to confirm, I could be wrong!!) but I'm wondering if a Restriction is what we want here? From the Mergify docs, "A Restriction defines a list of user or team for whom a restriction applies." so it sounds like it's a team/user that gets an extra set of rules applied to them?

The `require_code_owner_reviews` rule "blocks merging pull requests until code owners review them" sounds more like what we might be looking for? https://help.github.com/articles/about-codeowners/

fuzzball81:
I have dropped the restrictions for now. We can look at those types of protections later if we need too. The goal is to get things working then optimize from there.

PR #20, opened 2018-07-23
Fix formatting for doc strings

jpichon:
A function's description needs a blank line before listing the
parameters, or the line breaks won't be displayed in the module
content's documentation.

Also fix up a couple of inconsistent quote marks (''' vs """) that were
missed in PR#16.


jpichon:
Examples of the problem are visible at e.g. https://git-wrapper.readthedocs.io/en/latest/source/git_wrapper.html#git_wrapper.base.GitWrapperBase.add_remote , the parameters get all scrunched up into the one paragraph.

PR #22, opened 2018-07-24
Import 'mock' from the unittest library

jpichon:
This stops the 'make' commands related to testing/coverage and direct calls to py.test (as we document them in the CONTRIBUTING.rst file) from failing.

### DIFF
--- a/tests/test_rebase.py
+++ b/tests/test_rebase.py
@@ -1,6 +1,6 @@
 #! /usr/bin/env python
 
-from unittest.mock import patch
+from mock import patch
 
 import git
 import pytest


tests/conftest.py
```diff
@@ -1,7 +1,7 @@
 #! /usr/bin/env python
 """Base fixtures for unit tests"""
 
-from mock import Mock
+from unittest.mock import Mock
```
hguemar: Two options here:
1. use standalone mock module as openstack does
2. use six
```python
import six
six.add_moves(six.MovedModule('mock', 'mock', 'unittest.mock')
from six.moves import mock
from mock import Mock 
```

jpichon: Ah, I'd be happy with 1) too, that'd resolve the problem I mentioned. I guess we never added 'mock' explicitly to the requirements.txt to begin with, that'd be enough.

fuzzball81: Let’s go with the direct import of mock since it works in all cases and does not require six.

setup.py does have mock in the test requirements. I would prefer keeping requirements.txt for only run time dependencies.

hguemar: PR#25 adds a test-requirements.txt file where I explicitly add mock module so it should be fine then.

PR #23, opened 2018-07-24
Add some exception handling

jpichon:


jpichon:
Note: This PR depends on the logging one (PR#21), not quite sure how dependent PRs work in GitHub

jpichon:
I thought I would be able to make an inline comment but it doesn't look like it.

Because of the description for test_describe_empty at https://github.com/release-depot/git_wrapper/pull/23/commits/58f5ffdf31d575a040057d14405531eea564041e#diff-d06796e77090d59038f12fe9b1260baeR104
     """
     GIVEN GitWrapperBase initialized with a path and repo
     WHEN describe is called with a bad hash
     THEN a dictionary with an empty tag and patch is returned
    """
I was wondering if the preference may be to return an empty dictionary when the hash doesn't exist rather than raise an exception? If so, let me know and I'll update accordingly.

jguiditta:
I'll wait for @fuzzball81 to be back to weigh in, but my thought it this would make more sense to raise an error than to just return an empt dict. However, maybe there is a reason not to raise an exception that I am unaware of.

git_wrapper/base.py
```diff
@@ -69,7 +71,11 @@ def describe(self, sha):
            :return dict: A dict with tag and patch data
         """
         ret_data = {'tag': '', 'patch': ''}
-        output = self.git.describe(sha).split('-g')
+        try:
+            output = self.git.describe(sha).split('-g')
+        except git.CommandError as ex:
+            raise exceptions.ReferenceNotFoundException("Could not find sha %s." % sha) from ex
```
hguemar: This is clearly a situation where we want to raise an exception.
I'd be more comfortable if we added the --tags or --all options as it will always fail without annotated tags in repo.

fuzzball81: I don't have an issue with adding the '--all' option here to handle annotated tags.

jpichon:
Thanks for the comments, everyone. I added the --all flag as requested. I'm not sure what a test would look like though. From what I can tell the result is the same from the library's perspective. From the caller's perspective it seems to return the tag prefixed by tag/, so something to be aware of. Additionally the command still fails if a repo has no tags.

I realise my raising a ReferenceNotFound exception was too naive, when describe can fail in a variety of interesting ways. I poked a bit to see if we could be more specific but all errors return code 128, and the detail is only in an error message that comes from git itself and can be localised, so string parsing is a no-go. I generalised the exception for now. We could be smart and run a bunch of checks ourselves if we want the details (like trying to list the tags to see if a repo has any, using one of the search commit functions to guess if it was a problem with the hash, etc...), not sure if this is something we want right now?

fuzzball81:
If we are getting 'tag/blah' then we should strip that off before adding it to the dictionary to not break usage in other tooling also to avoid every caller having to strip that off in there own special way.

I think we need several types of exceptions, maybe a general one here works best since GitPython is doing a general system call to git describe. Since the return codes are all the same, then a catch all makes sense here.

jpichon:
Thank you for the comment, I'll update with additional stripping logic. Agreed on specific exceptions being nicer when we can get them!

fuzzball81:
Let go the general route for now. I am going to add an enhancement issue for looking into extracting failure details from describe add adding those to a describe exception.

PR #25, opened 2018-07-27
Add coverage in tests reports

hguemar:
Makes it easier to find where to add tests.
Use requirements files to handle tests dependencies.


hguemar:
@fuzzball81 this is a valid concern, I would suggest to use pbr which would simplify things by keeping dependencies in *requirements.txt files but that should be done in a separate review.

hguemar:
Rebase + add a new commit that migrates setup.py to pbr

setup.cfg
```diff
@@ -1,26 +1,32 @@
...
+packages =
+    git_wrapper
 
 [aliases]
-# Define setup.py command aliases here
 test = pytest
 
 [tool:pytest]
-collect_ignore = ['setup.py']
+addopts = tests
```
fuzzball81: Should this be test instead of tests?

hguemar: Nope, this is the tests directory (if I got your question right). By default, it looks inside the main package directory, but since it's outside, we need this  section to indicate that.

hguemar:
Fixed coverage reports by using pytest-cov to get consistent data.

setup.cfg
fuzzball81: That makes sense.

hguemar: This is handled in setup.py (the line with setup_requires=['pbr'])

PR #28, opened 2018-08-08
Fix rebasing on branch issue

jpichon:
Rebase currently creates a 'floating' commit as a result, as opposed to
rebasing the branch itself. This fixes it so that the branch gets
properly updated with the rebase result after running the command.


jpichon:
I don't think there is a reasonable unit test for this, with a mocked Repo object. It'll be a nice one for integration testing with a real git repo though.

The patch includes some additional cleanups but L49 is the only change necessary for fixing the bug.

PR #30, opened 2018-08-30
Skeleton code for integration tests

jpichon:
Also include a couple of tests & cleanup code to exercise it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM python:3.6
+
+# Clone an extra copy of the repo, to mess with during the tests
+RUN git clone https://github.com/release-depot/git_wrapper /tests/git_wrapper ; git config --global user.name "GitWrapper Integration" ; git config --global user.email "gitwrapper@example.com"
+
+# Copy the current state, to run the tests from
+WORKDIR /git_wrapper
+
+COPY . /git_wrapper
+
+RUN pip install -r requirements.txt -r test-requirements.txt
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -1,0 +1,23 @@
+import git
+import pytest
+
+
+REPO_ROOT = "/tests/git_wrapper"
+
+
+@pytest.fixture(scope="function")
+def repo_root():
+    # Check the starting point is sane
+    repo = git.Repo(REPO_ROOT)
+    origin = repo.remotes.origin
+
+    repo.heads.master.checkout()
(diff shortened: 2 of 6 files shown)


integration_tests/test_rebase.py
```diff
@@ -0,0 +1,48 @@
...
+
+    assert rebase.repo.active_branch.name == "master"
+    assert rebase.repo.head.object.hexsha != rebase_to
+
+    # Create a branch based on an old tag
+    rebase.repo.git.branch(branch_name, "0.0.1")
+
+    # Rebase that branch
+    rebase.to_hash(branch_name=branch_name, hash_=rebase_to)
+    assert rebase.repo.active_branch.name == branch_name
```
jpichon: That line tests for that fix in PR28, @fuzzball81 :)

fuzzball81: Great!

fuzzball81: Great!

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.6
+
+# Clone an extra copy of the repo, to mess with during the tests
+RUN git clone https://github.com/release-depot/git_wrapper /tests/git_wrapper
+
+RUN git config --global user.name "GitWrapper Integration"
+RUN git config --global user.email "gitwrapper@example.com"
```
fuzzball81: We should be able to combine all 3 of the run lines into one. This will only create one container layer for this part of the build and should help over all build/run time.

jpichon: Thanks, I didn't realise that! Will update.

integration_tests/conftest.py
```diff
@@ -0,0 +1,23 @@
+import git
+import pytest
+
+
+REPO_ROOT = "/tests/git_wrapper"
```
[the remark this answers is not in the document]
jpichon: Most container examples I found seem to use a path from root for building/testing their apps, and the Docker tutorial does as well. I'm not sure if it really matters, but I really don't have a strong opinion on this so, happy to switch to whatever is preferred!

jpichon:
I think it makes sense to use a container in this case, because we're messing with a bunch of things on the file system. The runtime cost doesn't seem too dire so far, just a minute when the other tests/flake8 take about 30-50 seconds to run at the moment. Maybe the tip @fuzzball81 shared in the review will help shave the time down a bit more too! I'll peek at the koji_wrapper docker stuff too, see if there are lessons learnt we could adapt here. Thanks.

PR #31, opened 2018-09-06
Add back support for Python 2.7

jpichon:
Consumers for the library (such as rdopkg) still need to support Python 2 for now and cannot use the library as it currently stands.

Note: Adding the wrapt dependency to simplify the Python 2 code for the decorator. Wrapt appears to be already packaged in the relevant places.


jpichon:
@jguiditta Thanks for the links! Looking through the patch again, with wrapt resolving the compatibility issues in the decorator, it looks like the only other use for six is for the "raise from" construct. One of the library recommended by your links is 'future' which seems to work the same way for this particular construct, but if word on the street is that it's more forward-looking friendly it looks like little work to switch it over (although I'm not sure the library is available downstream at this point? It is in Fedora.). We could also decide to lose the additional information from exception chaining for now, since a lot of it is present in the custom exception that's raised.

I'll upload a new version that uses python-future. I'll also see about making the wrapt import specific to Python 2 in the decorator, so it's easier to remove in the future.

PR #38, opened 2018-10-26
Minor string and logging improvements

jpichon:
Use str.format() where sensible, take advantage of logging deferred formatting in debug/info statements, centralise the logger setup.

### DIFF
--- a/git_wrapper/base.py
+++ b/git_wrapper/base.py
@@ -9,13 +9,10 @@ from future.utils import raise_from
 from git_wrapper import exceptions
 
 
-logger = logging.getLogger(__name__)
-
-
 class GitWrapperBase(object):
     """Provides a wrapper to interact with a git repository"""
 
-    def __init__(self, path='', repo=None):
+    def __init__(self, path='', repo=None, logger=None):
         """Constructor for GitUtilBase object
 
             :param str path: Path to a git repo Default('')
@@ -23,6 +20,13 @@ class GitWrapperBase(object):
         """
         self.__repo = None  # Added to clear pylint warnings
         self.__setup(path, repo)
+        self._setup_logger(logger)
+
+    def _setup_logger(self, logger):
+        if logger:
+            self.logger = logger
+        else:
+            self.logger = logging.getLogger(__name__)
 
     def __setup(self, path, repo):
         """Sets the path and repo after performing validation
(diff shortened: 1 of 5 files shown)


git_wrapper/base.py
```diff
@@ -75,7 +75,7 @@ def describe(self, sha):
         try:
             output = self.git.describe('--all', sha).split('-g')
         except git.CommandError as ex:
-            raise_from(exceptions.DescribeException("Error while running describe command on sha %s: %s" % (sha, ex)), ex)
+            raise_from(exceptions.DescribeException("Error while running describe command on sha {sha}: {error}".format(sha=sha, error=ex)), ex)
```
hguemar: PEP3101 <3

git_wrapper/clone.py
```diff
@@ -84,7 +84,7 @@ def destroy_and_reclone(self):
         for name, url in remotes.items():
             if name != default_remote:
                 try:
-                    logger.debug("Adding remote {remote}".format(remote=name))
+                    logger.debug("Adding remote %s", name)
```
hguemar: Not sure, why we're reverting to old string formatting?

jpichon: Note the use of a comma instead of the % operator. This defers string formatting until it's actually used, which is useful for debug/info level messages that may not be parsed at all.

git_wrapper/clone.py
```diff
@@ -12,17 +11,16 @@
...
 class GitWrapperClone(GitWrapperBase):
     """Provides git clone functionality"""
 
-    def __init__(self, path='', repo=None):
+    def __init__(self, path='', repo=None, logger=None):
         # If we're cloning for the first time, we may legitimately not
         # have a path or repo to use yet - so don't initialize just yet
         if path or repo:
-            super(GitWrapperClone, self).__init__(path=path, repo=repo)
+            super(GitWrapperClone, self).__init__(path=path, repo=repo, logger=None)
```
jguiditta: why not just pass in logger here, and let the super init it as needed?

git_wrapper/base.py
```diff
@@ -75,7 +79,7 @@ def describe(self, sha):
...
```
jguiditta: This line is extremely long, I can't even easily read it in the PR, can you wrap it when you get a chance?

git_wrapper/base.py
```diff
@@ -9,20 +9,24 @@
...
         """
         self.__repo = None  # Added to clear pylint warnings
         self.__setup(path, repo)
+        self._setup_logger(logger)
+
+    def _setup_logger(self, logger):
+        if logger:
+            self.logger = logger
+        else:
+            self.logger = logging.getLogger('git_wrapper')
```
jguiditta: Could we still use __name__ here, so the logs are attached to the calling object?

git_wrapper/clone.py
jpichon: Whoops! Typo, sorry about this.

git_wrapper/base.py
jpichon: Done.

git_wrapper/base.py
jpichon: Sure. Everything seemed to get associated with "git_wrapper.base" no matter what in my local tests, but I suppose that would be adjusted via a logging config anyway.

PR #39, opened 2018-11-07
Additional integration tests

jpichon:
Also clean up a bit more thoroughly between integration tests.

- -
We can leave this PR waiting as a WIP until the redesign is done, that is fine. I can refactor it then.

### DIFF
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -26,6 +26,11 @@ def repo_root():
     repo.heads.master.checkout()
     repo.head.reset(origin.refs.master, index=True, working_tree=True)
 
+    # Remove extra local branches
+    for b in repo.branches:
+        if b.name != 'master':
+            repo.git.branch("-D", b.name)
+
 
 @pytest.fixture(scope="function")
 def clone_repo_root():
@@ -39,3 +44,25 @@ def clone_repo_root():
 @pytest.fixture
 def clone_repo_url():
     yield CLONE_REPO_URL
+
+
+@pytest.fixture(scope="function")
+def patch_cleanup():
+    # Avoid state leaking into other tests in case of early failure
+    yield
+    repo = git.Repo(REPO_ROOT)
+    try:
+        repo.git.am('--abort')
+    except git.GitCommandError:
+        pass
+
+
+@pytest.fixture(scope="function")
+def rebase_cleanup():
+    # Avoid state leaking into other tests in case of early failure
+    yield
+    repo = git.Repo(REPO_ROOT)
+    try:
+        repo.git.rebase('--abort')
+    except git.GitCommandError:
+        pass
--- a/integration_tests/test_commit.py
+++ b/integration_tests/test_commit.py
@@ -1,0 +1,13 @@
+from git_wrapper.repo import GitRepo
+
+
+def test_revert(repo_root):
+    repo = GitRepo(repo_root)
+    repo.git.checkout("0.1.0")
+    assert "history" in repo.repo.head.object.message
+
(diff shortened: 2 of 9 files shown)


integration_tests/conftest.py
```diff
@@ -9,6 +9,26 @@
 REPO_ROOT = "/tests/git_wrapper"  # Local copy
 
 
+def clean_up_ongoing_operations(repo):
```
fuzzball81: I would suggest passing the operation to clean up 'am', 'rebase' and then calling the correct abort from that point instead of using fall through logic.

integration_tests/test_commit.py
```diff
@@ -0,0 +1,142 @@
...
+
+from git_wrapper import exceptions
+from git_wrapper.commit import GitWrapperCommit
+
+
+def test_apply_diff(repo_root):
+    commit = GitWrapperCommit(repo_root)
+    test_branch = "test_apply_diff"
+
+    # Create a diff file
```
fuzzball81: I would consider pulling this off to a data file and then read it in at test time. You could us something like pytest_datadir (https://github.com/gabrielcnr/pytest-datadir) to easily find the file. We use pytest-datadir in atkinson and koji-wrapper for testing.

integration_tests/test_commit.py
```diff
@@ -0,0 +1,142 @@
...
+    assert "Test commit message" in message
+    assert "Signed-off-by" in message
+
+
+def test_apply_patch(repo_root):
+    commit = GitWrapperCommit(repo_root)
+    test_branch = "test_apply_patch"
+
+    # Create patch file (based on git format-patch)
+    patch_path = "./test.patch"
```
fuzzball81: Same comment as the other patch text above.

integration_tests/test_commit.py
```diff
@@ -0,0 +1,142 @@
...
+    assert "Test patch" in commit.repo.head.object.message
+
+
+def test_abort(repo_root):
+    commit = GitWrapperCommit(repo_root)
+    test_branch = "test_abort"
+
+    # Create a bad patch file
+    patch_path = "./test.patch"
+    patch = """From 45e320cfcbf02a5e8bc3cd607f90949b0f5d0617 Mon Sep 17 00:00:00 2001
```
fuzzball81: Same comment as the other patch text above.

integration_tests/conftest.py
jguiditta: While I am all for making code cleaner, I would strong advocate NOT making such a fix if it means making it herder to write tests - it is often hard enough to convince contributors to add them, we should keep that bar as low as possible, imo.  I am sure we can decompose this in some way to clean it up, but please, just not in a way that makes testing more onerous.

jpichon: Good point! What I have in mind is referencing an additional (existing) fixture argument in the test def, so not particularly onerous but something to remember when writing the test or as reviewers. It seems to work so I'll run with it and we can discuss in the next review iteration once it's clearer what that looks like. Personally I find addCleanup()/addFinalizer()-type construct more natural and easier to understand but they only seems to work within fixtures, not within tests directly.

jpichon:
> One thing to consider here is what file these tests are trying to patch. We may want to have a dummy file as part of the tests that can be used in place of using one of the docs files. I could see us having problems in the future is that file need to change for some reason.

That's why every test starts by creating a branch off an old commit. Even if we rearchitecture everything and remove/rename that file in the future, it will still be present in the expected state on the branch or checkout.

jpichon:
(Rebased)

Issue #36, opened 2018-10-16
Rethink the architecture/design


jpichon:
* git_wrapper version: 0.1.0

Cf. discussion on PR #34 :

>  I do think we might want to consider the design direction we are going in here though. We are starting to get a lot of classes that are all actions, vs representing actual data objects. To me, cloning is an action that you take on a repo object, and so feels more like a method on some other object, rather than a standalone object in and of itself. I am not blocking this in any way, but I think we should give some thought before venturing too much further down this path, as I am not sure this approach is going to scale as well as we would hope.

> I would recommend that we take some of these actions and turn them into function, while others turn into object method. I do want to avoid making one object to rule them all. That is an anti-pattern in itself (https://en.wikipedia.org/wiki/God_object)

> Agreed on all points. I was certainly not suggesting we move everything into a single mega-object, which I think you realize, but it is a good anti-pattern to beware of nonetheless.

jpichon:
@jguiditta Hi! Would you mind doing a braindump here of the ideas you have you mentioned in the PR, so we have some food for thought to get started with? (I'm a bit short!) It'd be awesome if we could handle the design improvements within the next couple of weeks, and moved on to writing other missing features after that. Thank you.

fuzzball81:
The thought that I had is to break down parts of a git working tree into objects. This would mean creating one main repo object that allows us to access the other objects like branches, remotes, etc. That means many of the classes we have now would then be methods on these other objects. Things like clone would be a method on the repo object and rebase would move to the branch object. This approach is more object composition based, where several objects work together to form a whole.

For functionality that does not really belong in one of these objects, things like cherry or crosses several objects, we can make then functions that take in the objects needed at parameters.

This is just my current idea, I am interested in hearing with others are thinking.

jguiditta:
That is basically what I had meant as well.  I started going through the current codebase yesterday to see what it was doing overall, and so far have seen objects like Repo, Commit, Log - and then moving existing classes into methods where appropriate on those (and probably some other) objects, as you mention above.  I will continue to look at this today and post some more detailed thoughts on this as soon as I can.

fuzzball81:
@jpichon @jguiditta I see two paths. 
[...]

PR #40, opened 2018-11-16
Redesign 1: GitWrapperBase becomes GitRepo

jpichon:
I think it'd be better to wait for the whole thing to be ready before merging as the structure is a bit weird while it's halfway, but posting this as a PR to make it easier to see what it looks like and discuss/share early feedback.

See related issue #36.

### DIFF
--- a/git_wrapper/repo.py
+++ b/git_wrapper/repo.py
@@ -9,11 +9,11 @@ from future.utils import raise_from
 from git_wrapper import exceptions
 
 
-class GitWrapperBase(object):
+class GitRepo(object):
     """Provides a wrapper to interact with a git repository"""
 
     def __init__(self, path='', repo=None, logger=None):
-        """Constructor for GitUtilBase object
+        """Constructor for GitRepo object
 
             :param str path: Path to a git repo Default('')
             :param git.Repo repo: An already constructed git.Repo object to use Default(None)
--- a/git_wrapper/cherry.py
+++ b/git_wrapper/cherry.py
@@ -3,10 +3,10 @@
 
 import re
 
-from git_wrapper.base import GitWrapperBase
+from git_wrapper.repo import GitRepo
 
 
-class GitWrapperCherry(GitWrapperBase):
+class GitWrapperCherry(GitRepo):
     """Provides git cherry functionality"""
 
     def __init__(self, *args, **kwargs):
--- a/git_wrapper/clone.py
+++ b/git_wrapper/clone.py
@@ -7,11 +7,11 @@ import shutil
 import git
 from future.utils import raise_from
 
-from git_wrapper.base import GitWrapperBase
+from git_wrapper.repo import GitRepo
 from git_wrapper import exceptions
 
 
-class GitWrapperClone(GitWrapperBase):
+class GitWrapperClone(GitRepo):
     """Provides git clone functionality"""
 
     def __init__(self, path='', repo=None, logger=None):
--- a/git_wrapper/commit.py
+++ b/git_wrapper/commit.py
@@ -6,12 +6,12 @@ import os
 import git
 from future.utils import raise_from
 
-from git_wrapper.base import GitWrapperBase
+from git_wrapper.repo import GitRepo
 from git_wrapper import exceptions
 from git_wrapper.utils.decorators import reference_exists
 
 
-class GitWrapperCommit(GitWrapperBase):
+class GitWrapperCommit(GitRepo):
     """Provides git commit functionality"""
 
     def _expand_file_path(self, path):
--- a/git_wrapper/rebase.py
+++ b/git_wrapper/rebase.py
@@ -4,12 +4,12 @@
 import git
 from future.utils import raise_from
 
-from git_wrapper.base import GitWrapperBase
+from git_wrapper.repo import GitRepo
 from git_wrapper import exceptions
 from git_wrapper.utils.decorators import reference_exists
 
 
-class GitWrapperRebase(GitWrapperBase):
+class GitWrapperRebase(GitRepo):
     """Provides git rebase functionality"""
 
     @reference_exists("branch_name")
--- a/tests/test_repo.py
+++ b/tests/test_repo.py
@@ -1,5 +1,5 @@
 #! /usr/bin/env python
-"""Tests for GitWrapperBase"""
+"""Tests for GitRepo"""
 
 from mock import Mock, patch
 
@@ -7,7 +7,7 @@ import pytest
 
 from git import CommandError
 
-from git_wrapper.base import GitWrapperBase
+from git_wrapper.repo import GitRepo
 from git_wrapper import exceptions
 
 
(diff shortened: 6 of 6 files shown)


jpichon:
> This looks like good progress, though I have to admit I found it a bit hard to take in due to the scope. You may be too far along to consider this, but as I was reading through, it occurred to me that it might be possible to do this in stages if we were to start with simply renaming GitWrapperBase to GitRepo, and updating all references. Then, assuming tests pass, change it over to some of the new design ideas one chunk at a time - we would not do a release until all that was done, but I don't see why the work could not be done incrementally, and it would make it much easier to parse the code submitted for reviews.

Thanks for all the comments! May I ask, are you looking at the single view with all the changes? I've been doing what you suggest with the individual commits to make it easier to take in (the first one is exactly that name change), and the tests pass for every individual commit.

I'm less sure about merging as we go, because I keep finding issues that require modifying things at the beginning again and I suspect it may end up even messier if we start merging now. Perhaps I just need to get a little bit further along first to iron out the last few larger issues. Thanks again for the comments, they are helpful.

git_wrapper/branch.py
```diff
@@ -42,8 +48,8 @@ def to_hash(self, branch_name, hash_):
 
         self.logger.debug("Successfully rebased branch %s to %s", branch_name, hash_)
 
-    def abort(self):
-        """Abort a rebase."""
+    def rebase_cancel(self):
```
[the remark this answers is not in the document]
jpichon: Putting the verb first does sound better! I was thinking about trying to make it easier to remember how to cancel operations with a single word, because when I was working on the commit.py stuff there were 3 different verbs to cancel an operation ("abort" applying a patch, "revert" a commit, "reverse" applying a diff) which gets confusing. Probably it is better to stick to the proper git terminology though, to avoid another kind of confusion.

git_wrapper/remote.py
```diff
@@ -0,0 +1,47 @@
+#! /usr/bin/env python
+"""This module acts as an interface for acting on git remotes"""
+
+import git
+
+
+class GitRemote(object):
```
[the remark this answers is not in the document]
fuzzball81: I don't think we need to do that here. Not sure what functionality from GitRepo we would need. I see this more of a building block class then an child class.

git_wrapper/repo.py
```diff
@@ -0,0 +1,185 @@
...
+        for name, url in remotes.items():
+            if name != default_remote:
+                try:
+                    self.logger.debug("Adding remote %s", name)
+                    r = self.repo.create_remote(name, url)
+                except git.GitCommandError as ex:
+                    msg = "Issue with recreating remote {remote}. Error: {error}".format(remote=name, error=ex)
+                    raise_from(exceptions.RemoteException(msg), ex)
+
+    def describe(self, sha):
```
fuzzball81: This should probably move to the GitCommit class or out to a utility function.

git_wrapper/repo.py
```diff
@@ -0,0 +1,185 @@
...
+
+import git
+from future.utils import raise_from
+
+from git_wrapper.branch import GitBranch
+from git_wrapper import exceptions
+from git_wrapper.remote import GitRemote
+
+
+class GitRepo(object):
```
fuzzball81: I wonder if we should consider using this class more like a proxy to access things like branches, remotes, commits etc instead of using it a as a base.

git_wrapper/branch.py
```diff
@@ -1,20 +1,26 @@
...
 from future.utils import raise_from
 
-from git_wrapper.base import GitWrapperBase
 from git_wrapper import exceptions
 from git_wrapper.utils.decorators import reference_exists
 
 
-class GitWrapperRebase(GitWrapperBase):
-    """Provides git rebase functionality"""
+class GitBranch(object):
```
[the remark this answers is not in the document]
fuzzball81: I don't think we need to do that here. Not sure what functionality from GitRepo we would need. I see this more of a building block class then an child class.

jguiditta:
> Thanks for all the comments! May I ask, are you looking at the single view with all the changes? I've been doing what you suggest with the individual commits to make it easier to take in (the first one is exactly that name change), and the tests pass for every individual commit.
> 

Yes, I did not click through the individual commits, I just went to the 'files changed' tab.  It may have been easier to read if I had looked at the individual commits, I will try again that way tomorrow.

> I'm less sure about merging as we go, because I keep finding issues that require modifying things at the beginning again and I suspect it may end up even messier if we start merging now. Perhaps I just need to get a little bit further along first to iron out the last few larger issues. Thanks again for the comments, they are helpful.

Again, you may be too far along to do that anyway, but in general, I am not a fan of having a sweeping redesign off in a separate branch and then merged in all at once. IMHO, it is easier to get a sense of where it is going, and spot problems earlier on, if the work is done incrementally.  I understand not everyone feels this way, I am just explaining my thought process a bit.

git_wrapper/repo.py
jpichon: Yep! I just haven't looked at it yet. I'm following the outline from issue #36 at this point, https://github.com/release-depot/git_wrapper/issues/36#issuecomment-436973211

jpichon:
> Again, you may be too far along to do that anyway, but in general, I am not a fan of having a sweeping redesign off in a separate branch and then merged in all at once. IMHO, it is easier to get a sense of where it is going, and spot problems earlier on, if the work is done incrementally. I understand not everyone feels this way, I am just explaining my thought process a bit.

Yes, I understand. I wish GitHub PRs worked a bit more like Gerrit where patches in a series can still be reviewed/approved individually, and the first few can merge even if there are issues with the later ones. I'm trying to keep the individual commits small, self-contained and independently reviewable. I don't mind having to rework future patches as comments come up; I also think some things are easier to reason about when they're laid out in front of you.

@jguiditta I could try to create multiple branches on my side and have one PR per commit, and number them so they can be reviewed in order. Do you think that would help?

jguiditta:
> Yes, I understand. I wish GitHub PRs worked a bit more like Gerrit where patches in a series can still be reviewed/approved individually, and the first few can merge even if there are issues with the later ones. I'm trying to keep the individual commits small, self-contained and independently reviewable. I don't mind having to rework future patches as comments come up; I also think some things are easier to reason about when they're laid out in front of you.
> 
Agree, the patch series concept from gerrit is a good one.

> @jguiditta I could try to create multiple branches on my side and have one PR per commit, and number them so they can be reviewed in order. Do you think that would help?

Hmm, I kind of like the idea, but I fear it would make the work cumbersome, which I certainly do not want - it is enough work already without adding unneeded manual steps on top (rebase of all those branches every time you change an earlier branch, for example).  I will leave it up to your discretion.  If you feel it would be a major benefit to go this route, and the overhead would not be as bad as I fear, go for it.  Otherwise, I can certainly just review one commit at a time in this PR.

jpichon:
> Hmm, I kind of like the idea, but I fear it would make the work cumbersome, which I certainly do not want - it is enough work already without adding unneeded manual steps on top (rebase of all those branches every time you change an earlier branch, for example). I will leave it up to your discretion. If you feel it would be a major benefit to go this route, and the overhead would not be as bad as I fear, go for it. Otherwise, I can certainly just review one commit at a time in this PR.

I will give it a shot. Nothing can really move until the redesign is completed and there's nothing more I can do at this point so if this can make reviewers' life easier to get it merged faster, I'm all for it.

jpichon:
Ok, done! You'll still have to use the "Commits" tabs to be able to review one/the latest commit only at a time on the later PRs. Someone who'd rather get a feel for how the whole lot looks together can use Redesign 9. If it turns out not to be all that helpful to reviewers after all we can close the other PRs and go back to a single one. In the meantime let's make sure they don't merge out of order and approve only one at a time!! :)

git_wrapper/repo.py
fuzzball81: No sure what that would look like at this point. Let see where this redesign takes us.

PR #41, opened 2018-11-27
Redesign 2: Move Remote functions into their own class

jpichon:


git_wrapper/repo.py
```diff
@@ -92,27 +88,17 @@ def describe(self, sha):
...
-            self.logger.debug("Failed to update new remote %s (%s), removing it. Error: %s", name, url, ex)
-            self.repo.delete_remote(remote)
+    @remote.setter
+    def remote(self, new_remote):
+        """Set up object to interact with Remotes
 
-        return ret_status
+            :param git_wrapper.remote.GitRemote new_remote: An already constructed GitRemote object to use
+        """
+        self._remote = new_remote
```
fuzzball81: We should use a type check here to make sure we get a GitRemote instance and not some other object. We could then raise a type error if we get something unexpected.

PR #44, opened 2018-11-27
Redesign 5: Move cherry functions into GitBranch

jpichon:


git_wrapper/branch.py
```diff
@@ -19,6 +21,28 @@ def __init__(self, git_repo, logger):
...
+        args = ['-v', upstream, head]
+        ret_data = {}
+        for line in self.git_repo.git.cherry(*args).split('\n'):
+            match = regex.match(line)
+            if match is not None:
+                ret_data[match.group(1)] = match.group(2)
+        return ret_data
+
+    def cherry_on_head_only(self, upstream, head):
+        """Get new patches between upstream and head"""
```
jguiditta: Minor doc fix on this and the next one, please include params list in doc

jpichon: Well spotted! If that's all right I made a note to do this after this merges, as it's outside the scope of the redesign. I just moved what was already there.

jguiditta: Yes, not a problem

PR #45, opened 2018-11-27
Redesign 6: Rename GitWrapperCommit into GitCommit

jpichon:


git_wrapper/commit.py
```diff
@@ -77,23 +84,23 @@ def apply_diff(self, branch_name, diff_path, message, signoff=False):
            :param bool signoff: Whether to add signed-off-by to commit message
         """
         # Ensure we don't commit more than we mean to
-        if self.repo.is_dirty():
-            msg = "Repository {repo} contains uncommitted changes. Please clean workspace before proceeding.".format(repo=self.repo.working_dir)
+        if self.git_repo.repo.is_dirty():
+            msg = "Repository {repo} contains uncommitted changes. Please clean workspace before proceeding.".format(repo=self.git_repo.repo.working_dir)
```
jguiditta: Linewrap, if you would

git_wrapper/repo.py
```diff
@@ -184,3 +186,18 @@ def branch(self, new_branch):
             :param git_wrapper.branch.GitBranch new_branch: An already constructed GitBranch object to use
         """
         self._branch = new_branch
+
+    @property
+    def commit(self):
+        """Return object to act on the repo's commits"""
+        if not self._commit:
+            self._commit = GitCommit(git_repo=self, logger=self.logger)
```
jguiditta: Same question about setter as in other patches

tests/test_commit.py
```diff
@@ -1,382 +1,382 @@
...
 def test_apply_patch_not_a_file(mock_repo):
     """
-    GIVEN GitWrapperCommit initialized with a path and repo
+    GIVEN GitRepo initialized with a path and repo
     WHEN apply_patch is called with a valid branch_name and invalid path
     THEN FileDoesntExistException raised
     AND git.am not called
     """
-    commit = GitWrapperCommit('./', mock_repo)
+    commit = GitRepo('./', mock_repo)
```
jguiditta: Minor thought here that did not occur to me in earlier reviews.  What if you made this the actual commit object contained in the GitRepo object as you had before?  Then you could simply call commit.action below instead of commit,commit.action. Maybe not worth it, and certainly not required for this patchset, just a thought - I am guessing there is some reason you switched, and I am just not getting it on a Friday afternoon...

jpichon: My thoughts are that from an end-user perspective it makes more sense to have your generic repo object and call what you need as you need it, rather than expect in advance you'll only need a subset, and be stuck when actually you need to do something else too. I changed the tests to reflect this usage pattern as well.

Having said that commit.commit() is the naming I'm least happy with, but that's easy to deprecate and change later if we decide commit.create() or something else is more sensible.

PR #46, opened 2018-11-27
Redesign 7: Move apply and reverse functions to GitBranch

jpichon:


git_wrapper/branch.py
```diff
@@ -21,6 +22,12 @@ def __init__(self, git_repo, logger):
         self.git_repo = git_repo
         self.logger = logger
 
+    def _expand_file_path(self, path):
```
jguiditta: This doesn't have to be now, but this seems generically useful for our projects, so it might make sense to move it to toolchest after this set of reviews.

git_wrapper/branch.py
```diff
@@ -80,3 +87,84 @@ def abort_rebase(self):
...
+        """Apply a diff on top of the specified branch.
+
+           :param str branch_name: The name of the branch or reference to apply the diff to
+           :param str diff_path: Path to the diff file
+           :param str message: Commit message
+           :param bool signoff: Whether to add signed-off-by to commit message
+        """
+        # Ensure we don't commit more than we mean to
+        if self.git_repo.repo.is_dirty():
+            msg = "Repository {repo} contains uncommitted changes. Please clean workspace before proceeding.".format(repo=self.git_repo.repo.working_dir)
```
jguiditta: linewrap

git_wrapper/branch.py
jpichon: Agreed, it's pretty clear to me that'll move to a utils type module at some point, or the tool chest may make sense too if it's needed elsewhere. Just not rushing it until it's actually needed.

git_wrapper/branch.py
jguiditta: I would vote for pep8 standard, it is easier to read with you have side by side panes in tmux/vim/whatever

jpichon: I'm somewhat partial to the pep8 standard as well, as I usually have an open file on a half screen and a terminal on the other half.

PR #47, opened 2018-11-27
Redesign 8: Move describe to GitCommit

jpichon:


tests/test_commit.py
```diff
@@ -10,6 +10,80 @@
...
+    WHEN commit.describe is called a tag value only
+    THEN a dictionary with a tag and empty patch is returned
+    """
+    expected = {'tag': '1.0.0', 'patch': ''}
+    attrs = {'describe.return_value': '1.0.0'}
+    mock_repo.git.configure_mock(**attrs)
+
+    repo = GitRepo('./', mock_repo)
+
+    assert expected == repo.commit.describe('12345')
```
jguiditta: should this be 1.0.0?

jguiditta: Ok, we can leave it - I was reviewing everything as new, and this one didn't seem right when I read it.

PR #49, opened 2018-12-04
Removing python 3.5 tests.

fuzzball81:
Python 3.6 seems to be what most distros are moving toward. Dropping the
python 3.5 setup in tox.


jpichon:
Agreed, LGTM.

PR #51, opened 2018-12-07
Log diff functionality

jpichon:


git_wrapper/branch.py
```diff
@@ -190,3 +190,21 @@ def reverse_diff(self, diff_path):
...
+
+           :param str hash_from: A commit hash
+           :param str hash_to: A commit hash
+        """
+        range_ = "{0}...{1}".format(hash_from, hash_to)
+        log = self.git_repo.git.log("--abbrev-commit",
+                                    "--pretty=oneline",
+                                    range_)
+
+        if log:
```
fuzzball81: One thing we may want to consider is the ability to not include the commit hash in the output. That could be an optional parameter that log_diff takes and we default it to return things as you have here.

fuzzball81: Upon deeper thinking. At the end of the day we want the log data to come out mostly raw. I am thinking of the rdopkg case where searching for rhbz tags require the entire log message to be available. We have a few options to handle this.

1) We have methods to handle post-formatting of the data. For example the method here would be a one_line method. This would allow a user to consume the log data based on some pre-defined formats. Others may need to be added later to support other use cases.

2) Allow for an emitter/format function to be passed into this call. If one is not provided all of the log data is returned. This would put the formatting of the data in the hands of the user so this one call can support many use cases with no code change.

I like the pattern option 2 presents. This library is then not making decisions on how the data is presented instead it focuses on getting the data to the user to work with.

I am interested to here others thoughts.

jpichon: Something like having a list of pre-defined formatter keywords ({commit} {abbrev-commit} {shortlog} {etc}?) and letting the caller go mad with it? That does sound cool. Error handling should be interesting :-) I wonder if we may want to offer a couple of sane defaults, like we have with cherry: a main function (but not private in this case) and a couple of other functions like short_log_diff or full_log_diff that call to it with a predefined formatter, so people don't have to figure out a pattern every time for the most common cases.

fuzzball81: I like this idea. Having the custom format emitter and then some predefined ones sounds like a good approach. We can deal with the error checking portion by taking the format from the using, converting it to a template string thing using [safe_substitute](https://docs.python.org/3/library/string.html#string.Template.safe_substitute). If the data is not present or unavailable we still return what we can. This would mean breaking down the log data in something like a dictionary with commonly known keys, but that would be simple.

All of this makes me wonder if the initial data that is returned in the call here is a log object that has all of this formatting capability.

jguiditta: I think I would lean toward the idea of a log object type, at least in theory.  However, we did say in the redesign thread that we would keep the number of objects small to start and break them out as needed, so maybe see how this develops before taking that step?  Also, I think the idea of a formatter (which, @fuzzball81 I believe you just added somewhere as well, but I can't remember where).  Another option would be to simply pass in a config hash with the settings you want, and our code parses that into options to pass to git.  So, for the example above, you might pass in 
```
config = {"abbrev-commit": None,
               "pretty": "oneline"}
# pseudcode to parse:
def _parse_config(config):
    return [f"--{k} {v}" for k,v in config]
```
Then, if any invalid option are passed in, bubble up the error.

jguiditta: > Something like having a list of pre-defined formatter keywords ({commit} {abbrev-commit} {shortlog} {etc}?) and letting the caller go mad with it? That does sound cool. Error handling should be interesting :-) I wonder if we may want to offer a couple of sane defaults, like we have with cherry: a main function (but not private in this case) and a couple of other functions like short_log_diff or full_log_diff that call to it with a predefined formatter, so people don't have to figure out a pattern every time for the most common cases.

I also like the idea of a few simple, predefined wrappers being included for common cases, to simply client code

jguiditta: So, I think my net opinion, summing up the above comments, would be to make the existing log_diff method have a slightly different signature, taking a hash or formatter, and later decide if there is enough about logs to warrant an object encapsulating those concepts

jpichon:
Not entirely sure GitHub sends notifications on push updates - I think this solution synthesises all the comments in a decent compromise so that everyone is equally unhappy ;) Thanks for the pointer to safe_substitute!

PR #54, opened 2018-12-20
Add functionality to refresh a repository

jpichon:


git_wrapper/repo.py
```diff
@@ -189,3 +189,90 @@ def commit(self, new_commit):
...
+           :param str remote: Remote to sync
+        """
+        try:
+            remote = self.repo.remote(remote)
+        except ValueError:
+            msg = "Remote {remote} does not exist on repo {repo}".format(
+                remote=remote,
+                repo=self.repo.working_dir
+            )
+            raise exceptions.ReferenceNotFoundException(msg)
```
fuzzball81: I like this check, I am wondering if we should pull this out like reference_exists and make it a decorator. The other idea would be make it a utility that library users could call to check to see if a remote exists in the repo.

git_wrapper/repo.py
```diff
@@ -189,3 +189,90 @@ def commit(self, new_commit):
...
+                self.sync(remote)
+            except exceptions.RemoteException:
+                self.logger.exception("Error fetching remote %s", remote)
+                errors.append(remote)
+
+        if errors:
+            msg = "Error fetching these remotes: {0}".format(", ".join(errors))
+            raise exceptions.RemoteException(msg)
+
+    def reset(self, branch="master", remote="origin", remote_branch="master",
```
fuzzball81: On thought here is to allow for a 'type' parameter the defaults to hard. We can then extend this to allow for a soft reset if we have a need later. Otherwise I would recommend making it clear this is a hard reset by changing the name.

git_wrapper/repo.py
jpichon: Cool! I think I would rather wait until it's needed elsewhere before extracting it for now. Maybe we'll understand better if the utility way or the decorator way makes more sense then?

git_wrapper/repo.py
jpichon: Good point! I'll rename.

I'm also realising that although I was thinking about this in terms of "refreshing a repo", the sync() methods actually would fit better under "remote", while reset() applies to a "branch" so I'll move things around a bit in the next patch set.

git_wrapper/repo.py
fuzzball81: That's fair. If we find more need in other methods then we can pull it out then.

PR #58, opened 2019-01-18
Prep for version 0.2.0 release.

fuzzball81:

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -14,3 +14,14 @@ History
 * Base functionality
 * git cherry support
 * Removing pipenv support.
+
+0.2.0 (2019-01-18)
+------------------
+
+* Third release on PyPI.
+* Major refactor to better align the data model.
+* Added several new features
+        * Log diff generation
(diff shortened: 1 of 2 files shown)


HISTORY.rst
```diff
@@ -14,3 +14,14 @@ History
...
+0.2.0 (2019-01-18)
+------------------
+
+* Third release on PyPI.
+* Major refactor to better align the data model.
+* Added several new features
+  * Log diff generation
+  * Improved clone support including bare repos and destroy/reclone
+  * Revert support
+  * Support for rebasing to a branch or commit.
```
jpichon: So sorry to end up nit-picking, it looks like the sub-list isn't displaying properly: https://github.com/release-depot/git_wrapper/blob/4e434525f4058696b0d406cb5fd2ba4e2e03a663/HISTORY.rst . I think perhaps an extra space is needed? 😰

fuzzball81: Looks like it needed a tab character, should be fixed now. Good catch!

PR #60, opened 2019-02-20
Update mergify config for v2 api.

jguiditta:


jpichon:
Looks like manual merges are required for mergify config changes

PR #66, opened 2020-07-21
Update README with new release instructions

jpichon:


jpichon:
I don't have any strong thoughts either, happy to move it if that's preferred. I put it where I thought I'd be most likely to find it again in a year ;)

fuzzball81:
I am fine with either path as well. I believe you have to have the correct permissions on PYPI to upload a new version. I agree with @jpichon keeping it some place easy to find may be a good idea since doing a release is rare at the present time.

PR #70, opened 2021-01-06
Preserve square brackets in commit messages when applying patches

jpichon:
Commit messages in patches can include useful information between square brackets such as a bug/task reference or downstream-only status. The default git-am command strips those out but it is useful information users may want to preserve; this patch adds a parameter to enable doing that without changing the existing behaviour.


fuzzball81 (commented):
LGTM

jguiditta (approved):

PR #71, opened 2021-01-07
Add function to grep logs

jpichon:
Also move the existing log functions into a new GitLog component.

To help with reviewing this PR is broken down into two commits:
- The first one moves the existing log functions into a GitLog component. This only moves the existing code around, there are no functional changes to the contents.
- The second commit adds the new grep functionality and associated tests.

### DIFF
--- a/git_wrapper/repo.py
+++ b/git_wrapper/repo.py
@@ -11,6 +11,7 @@ from future.utils import raise_from
 from git_wrapper.branch import GitBranch
 from git_wrapper.commit import GitCommit
 from git_wrapper import exceptions
+from git_wrapper.log import GitLog
 from git_wrapper.remote import GitRemote
 from git_wrapper.tag import GitTag
 
@@ -28,6 +29,7 @@ class GitRepo(object):
         self.__repo = None  # Added to clear pylint warnings
         self._branch = None
         self._commit = None
+        self._log = None
         self._remote = None
         self._tag = None
 
@@ -212,3 +214,20 @@ class GitRepo(object):
         if not isinstance(new_tag, GitTag):
             raise TypeError("Tag must be a GitTag object.")
         self._tag = new_tag
+
+    @property
+    def log(self):
+        """Return object to act on the repo's logs"""
+        if not self._log:
+            self._log = GitLog(git_repo=self, logger=self.logger)
+        return self._log
+
+    @log.setter
+    def log(self, new_log):
+        """Set up object to interact with Logs
+
+            :param git_wrapper.log.GitLog new_log: Pre-constructed GitLog object
+        """
+        if not isinstance(new_log, GitLog):
+            raise TypeError("Log must be a GitLog object.")
+        self._log = new_log
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ wheels/
 *.egg-info/
 .installed.cfg
 *.egg
+*~
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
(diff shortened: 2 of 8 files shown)


git_wrapper/repo.py
```diff
@@ -212,3 +214,20 @@ def tag(self, new_tag):
         if not isinstance(new_tag, GitTag):
             raise TypeError("Tag must be a GitTag object.")
         self._tag = new_tag
+
+    @property
+    def log(self):
+        """Return object to act on the repo's logs"""
+        if not self._log:
+            self._log = GitLog(git_repo=self, logger=self.logger)
```
jguiditta, line 222: This may be semantics, but shouldn't this call the setter to be consistent?

jguiditta (commented):
Overall, LGTM.  One minor question, but no major issues that I see.

git_wrapper/repo.py
jpichon: It's consistent with the other setters, though I see what you mean. I tried changing it but I think it makes the code look more surprising:
```
        if not self._log:
            self.log = GitLog(git_repo=self, logger=self.logger)
        return self._log
```
My first thought would be that this looks like a typo, so I would tend to prefer the way it's currently done for clarity. But I'm happy to change it if the consensus is that it would be better to use the setter anyway, since we do get an additional safety type check that way.

jpichon (commented):
Thank you for the review!

git_wrapper/repo.py
jguiditta: In further reflection, I think the real problem is setting the variable in the getter.  Properties should be initialized in either the init, or some function called _from_ the init.  Any tricky logic of what to do for initialization of the property is ok to go in the setter if it makes sense, but when we set inside a getter, it breaks the encapsulation of this pattern.  I am sorry that I did not catch this earlier.  Here are a couple articles on usage that I found, just as examples (or you can look at base.py in koji_wrapper, where I did something similar):
https://www.programiz.com/python-programming/property
https://www.python-course.eu/python3_properties.php

I have noticed some inconsistency in whether people use '__var' or 'var' in the init, so that part is more up to taste, though to me it is cleanest to use the public attribute everywhere but inside the getter/setter.

jpichon: I see, yes, initialising them all in the __init__ as opposed to None would make more sense, and then the getter can just get without worrying about whether the variable is already set. Do you mind if I make that change in a separate PR? That way I can fix all of the other properties at the same time, without increasing the scope for this particular PR too much.

jguiditta: Sure, I think that is totally reasonable.  Sorry to propose extra work!

jpichon: Haha it's all for the better!!

fuzzball81: Here is the PEP8 guild on naming https://pep8.org/#method-names-and-instance-variables note that the internal var should contain the underscore.

One other note, I believe this also needs to be used with python2 for dlrn. Please correct me if I am wrong.

jpichon: That's correct, we're still running the tests with Python 2.7 as well.

jguiditta: Agree on the underscore, I was not meaning to suggest the internal variable itself should not have it, simply how it gets set on init.

jpichon: Ah, yes I misunderstood the location; removing the underscore was a means to use the setter. I should have the init update ready to go once this is merged. Marking thread as resolved since we agreed to handle it in a separate PR.

jguiditta (approved):

PR #76, opened 2021-02-10
Provide new helper related to the log

cjeanner:
Until now, we couldn't extract a single commit from a git repository.

This patch exposes a new method, log_show_commit. It also extract the
commit templating and pushes it in a new public method "commit_format",
allowing to actually format any commit content, even from another
application.

It also reduces code duplication.

### DIFF
--- a/git_wrapper/log.py
+++ b/git_wrapper/log.py
@@ -18,10 +18,8 @@ class GitLog(object):
         self.git_repo = git_repo
         self.logger = logger
 
-    @reference_exists('hash_from')
-    @reference_exists('hash_to')
-    def log_diff(self, hash_from, hash_to, pattern="$full_message"):
-        """Return a list of strings for log entries between two hashes.
+    def commit_format(self, commits, pattern='$full_message'):
+        """Return a list of strings representing the commits
 
            Any of the following placeholders may be used in the pattern:
              * $hash The full commit hash
@@ -33,16 +31,10 @@ class GitLog(object):
              * $author Commit author
              * $date Date the commit was authored
 
-           :param str hash_from: A commit hash
-           :param str hash_to: A commit hash
+           :param list commits: A list of commits
            :param str pattern: Formatter containing any of the placeholders above
+           :return: list of strings
         """
-        range_ = "{0}..{1}".format(hash_from, hash_to)
-        commits = self.git_repo.repo.iter_commits(range_)
-        if not commits:
-            return []
-
-        # Prepare patterns
         author_tpl = Template("$name <$email>")
         full_message_tpl = Template(
             "commit $hash\nAuthor: $author\nDate: $date\n\n$message"
@@ -68,8 +60,34 @@ class GitLog(object):
 
             line = line_tpl.safe_substitute(placeholders)
             log.append(line)
-
         return log
+
+    @reference_exists('hash_from')
+    @reference_exists('hash_to')
+    def log_diff(self, hash_from, hash_to, pattern="$full_message"):
+        """Return a list of strings for log entries between two hashes.
+
+           Any of the following placeholders may be used in the pattern:
+             * $hash The full commit hash
+             * $short_hash The short commit hash, similar to --abbrev-commit
+             * $message The commit message
+             * $summary First line of the commit message
+             * $full_message Complete commit info with hash, author, message.
+               Similar to default "git log" ouput
+             * $author Commit author
+             * $date Date the commit was authored
+
+           :param str hash_from: A commit hash
+           :param str hash_to: A commit hash
+           :param str pattern: Formatter containing any of the placeholders above
+           :return: list of strings
+        """
+        range_ = "{0}..{1}".format(hash_from, hash_to)
+        commits = self.git_repo.repo.iter_commits(range_)
+        if not commits:
+            return []
+
+        return self.commit_format(commits, pattern)
 
     def short_log_diff(self, hash_from, hash_to):
         """Return a list of strings for log entries between two hashes.
@@ -82,17 +100,19 @@ class GitLog(object):
         return self.log_diff(hash_from, hash_to, "$short_hash $summary")
 
     @reference_exists('branch')
-    def grep_for_commits(self, branch, grep_for, reverse=False, path=None):
+    def grep_for_commits(self, branch, grep_for, reverse=False, path=None,
+                         log_format='format:%H'):
         """Returns a list of matching commits shas.
            :param str branch: which branch to grep on
            :param str grep_for: what to grep for
            :param bool reverse: whether to return in reversed order
            :param str path: path to limit the search to, optionally
-           :return: A list of commit ids
+           :param str log_format: log format output. Defaults to format:%H. Please refer to git-log documentation about PRETTY FORMATS
+           :return: A list of resulting commit matching the pattern.
         """
         commits = []
 
-        params = [branch, "--format=format:%H", "--grep=%s" % grep_for]
+        params = [branch, "--format=%s" % log_format, "--grep=%s" % grep_for]
         if reverse:
             params += ['--reverse']
         if path:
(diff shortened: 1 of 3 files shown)


cjeanner:
Note: the "log_format" method is implicitly tested via the tests using the "pattern" parameter. If needed, I can create dedicated unit tests - lemme just know.

jpichon (changes requested):
Really nice work refactoring while keeping the existing methods doing what they should!! I made a couple of minor comments and suggestions for improvements. This is looking very good, thank you!

git_wrapper/log.py
```diff
@@ -107,3 +127,25 @@ def grep_for_commits(self, branch, grep_for, reverse=False, path=None):
             commits = results.split("\n")
 
         return commits
+
+    def log_show_commit(self, commit_hash=None, pattern="$full_message"):
+        """Return a string representing the given commit.
```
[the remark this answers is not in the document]
cjeanner: right. In order to be consistent, we should indeed return only a string, not a list with one string. Will correct it shortly.

git_wrapper/log.py
```diff
@@ -107,3 +127,25 @@ def grep_for_commits(self, branch, grep_for, reverse=False, path=None):
             commits = results.split("\n")
 
         return commits
+
+    def log_show_commit(self, commit_hash=None, pattern="$full_message"):
```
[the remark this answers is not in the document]
cjeanner: Fair - I can indeed explicitly put the default as HEAD, and use the decorator. I checked the GitPython code in order to see what was returned if no hash/reference was passed ;).

git_wrapper/log.py
```diff
@@ -18,10 +18,8 @@ def __init__(self, git_repo, logger):
         self.git_repo = git_repo
         self.logger = logger
 
-    @reference_exists('hash_from')
-    @reference_exists('hash_to')
-    def log_diff(self, hash_from, hash_to, pattern="$full_message"):
-        """Return a list of strings for log entries between two hashes.
+    def commit_format(self, commits, pattern='$full_message'):
```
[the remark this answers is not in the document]
cjeanner: yeah - the diff was a bit weird, buuut.... well. it's working ;).

git_wrapper/log.py
```diff
@@ -82,17 +100,19 @@ def short_log_diff(self, hash_from, hash_to):
...
            :param bool reverse: whether to return in reversed order
            :param str path: path to limit the search to, optionally
-           :return: A list of commit ids
+           :param str log_format: log format output. Defaults to format:%H
+           :return: A list of resulting commit matching the pattern.
         """
         commits = []
 
-        params = [branch, "--format=format:%H", "--grep=%s" % grep_for]
+        params = [branch, "--format=%s" % log_format, "--grep=%s" % grep_for]
```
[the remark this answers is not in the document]
cjeanner: oh, right. was due to my tests using "--pretty=format:..."
Though... in fact, I can probably drop THAT change, especially due to the way the output is later parsed. In my use-case, I'll just do 2 calls: one with the _grep in order to get commit ID(s), and one with the new show_commit() so that I get the actual information.

tests/test_log.py
```diff
@@ -160,3 +160,54 @@ def test_log_grep_bad_path(mock_repo):
...
+    WHEN log.log_show_commit is called with an unknown commit hash
+    THEN an empty list is returned
+    """
+    mock_repo.commit.return_value = ''
+    repo = GitRepo(repo=mock_repo)
+
+    with patch('git.repo.fun.name_to_object'):
+        change = repo.log.log_show_commit('0010000000000000000')
+
+    assert change == []
```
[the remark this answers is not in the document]
cjeanner: hmmmm ok. Weird, it didn't crash in the tests at least. I'll do some more tweaking.

git_wrapper/log.py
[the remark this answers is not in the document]
cjeanner: note: keeping it, with a better docstring and consistency.

tests/test_log.py
```diff
@@ -160,3 +160,54 @@ def test_log_grep_bad_path(mock_repo):
     with patch('git.repo.fun.name_to_object'):
         with pytest.raises(exceptions.FileDoesntExistException):
             repo.log.grep_for_commits("test_branch", 'test', True, 'wrongpath')
+
+
+def test_log_show_commit(mock_repo, fake_commits):
+    """
+    GIVEN GitRepo is initialized with a path and repo
+    WHEN log.log_show_commit is called with correct parameters
+    THEN a list of string should be returned
```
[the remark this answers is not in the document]
cjeanner: (in fact it will be a single string - comment updated anyway)

git_wrapper/log.py
```diff
@@ -107,3 +127,26 @@ def grep_for_commits(self, branch, grep_for, reverse=False, path=None):
...
+             * $date Date the commit was authored
+
+           :param str commit_ref: A commit hash. Defaults to HEAD
+           :param str pattern: Formatter containing any of the placeholders above
+           :return: A string
+        """
+        result = self.git_repo.repo.commit(commit_ref)
+        if result:
+            return self.commit_format([result], pattern)[0]
+        return ''
```
jpichon, line 152: Can we ever hit that, though? No results means the commit_ref was wrong, but that's caught by the decorator now?

integration_tests/test_log.py
```diff
@@ -65,3 +65,20 @@ def test_log_grep_badpath(repo_root):
...
+        'Author: Jason Joyce <jjoyce@redhat.com>\n'
+        'Date: Tue Jun 12 14:21:45 2018 -0400\n\n'
+        'Initial commit of README.rst\n'
+    )
+
+
+def test_log_show_commit_wrong_commit(repo_root):
+    repo = GitRepo(repo_root)
+    with pytest.raises(exceptions.ReferenceNotFoundException):
+        repo.log.log_show_commit('foo')
```
jpichon, line 84: Nice! I appreciate the extra integration tests, they're good for confidence since we have to mock so much.

integration_tests/test_log.py
```diff
@@ -65,3 +65,20 @@ def test_log_grep_badpath(repo_root):
 
     with pytest.raises(exceptions.FileDoesntExistException):
         repo.log.grep_for_commits('master', "Initial", path="badpath")
+
+
+def test_log_show_commi(repo_root):
```
jpichon, line 70: (Nit) commi -> commit

tests/test_log.py
```diff
@@ -160,3 +160,54 @@ def test_log_grep_bad_path(mock_repo):
...
+        "Date: Wed Dec 05 10:36:19 2018 \n\n"
+        "This is a commit message (#1)\nWith some details."
+    )
+
+
+def test_log_show_commit_no_commit(mock_repo):
+    """
+    GIVEN GitRepo is initialized with a path and repo
+    WHEN log.log_show_commit is called with an unknown commit hash
+    THEN an empty list is returned
```
jpichon, line 189: This should be "empty string" after the changes, though as per my comment above I wonder if the mock shouldn't raise an exception instead now?

jpichon (commented):
One question inline, changes looking good to me otherwise! Especially those integration tests :)

git_wrapper/log.py
cjeanner: well... I did hit this in the unit tests (tox -e py27) actually..... maybe I did it wrong though? If you see why "test_log_show_commit_no_commit" didn't fail with the Exception....

integration_tests/test_log.py
cjeanner: dang.. lemme correct.

tests/test_log.py
cjeanner: hm... in fact... Due to the way I mock this: ```mock_repo.commit.return_value = ''``` it won't raise anything.... Maybe that's only possible in the integration test?

tests/test_log.py
```diff
@@ -160,3 +160,53 @@ def test_log_grep_bad_path(mock_repo):
...
```
jpichon, line 189: This should be "raised an exception" now :)

git_wrapper/log.py
```diff
@@ -107,3 +127,25 @@ def grep_for_commits(self, branch, grep_for, reverse=False, path=None):
...
+               Similar to default "git log" ouput
+             * $author Commit author
+             * $date Date the commit was authored
+
+           :param str commit_ref: A commit hash. Defaults to HEAD
+           :param str pattern: Formatter containing any of the placeholders above
+           :return: A string
+        """
+        result = self.git_repo.repo.commit(commit_ref)
+        if result:
```
jpichon, line 150: I think the if statement can be removed too then? We know there'll be a result thanks to the decorator check, and if somehow for a mysterious reason there isn't, it shouldn't be silently hidden.

jpichon (approved):
Great. Thanks for making the changes!

PR #78, opened 2021-02-17
Use Github Actions to run the tests and remove Travis

jpichon:
I'm not 100% about the Mergify change so we may want to be careful about approving before CI passes at first. From https://github.com/Mergifyio/mergify-engine/discussions/1995 the job names need to be spelled out for the entire matrix.


.github/workflows/test.yml
```diff
@@ -0,0 +1,53 @@
...
+          # Run tox using the version of Python in `PATH`
+          run: tox -epy
+    flake8:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v2
+        - name: Setup Python
+          uses: actions/setup-python@v2
+          with:
+            python-version: 3.9
```
jguiditta, line 33: So we don't want to run flake the matrix of versions?

.github/workflows/test.yml
```diff
@@ -0,0 +1,53 @@
...
+          with:
+            python-version: ${{ matrix.python }}
+        - name: Install Tox and any other packages
+          run: |
+            python -m pip install --upgrade pip
+            pip install tox tox-gh-actions
+        - name: Run Tox
+          # Run tox using the version of Python in `PATH`
+          run: tox -epy
+    flake8:
```
jguiditta, line 26: I think you probably know this, but just to be sure - I believe this sets up a separate container to run flake8.  This may not be an issue, but it is likely to make the tests run a bit slower.  I believe on the ones I changed, I just made flake8 a step, which would also run it for each of the pythons listed above.

fuzzball81:
We will also need to change the mergify config to not use travis anymore. Not sure if it has github action support or not.

fuzzball81:
Sigh it helps if I read the file names while looking at the changes. I see the mergify changes now.

.mergify.yml
```diff
@@ -7,5 +7,10 @@ pull_request_rules:
       conditions:
           - label!=no-mergify
           - '#approved-reviews-by>=1'
-          - status-success=continuous-integration/travis-ci/pr
```
jguiditta, line 10: I think you should be able to just make this one line, using:
```
- status-success=Run tests
```

I do not believe you need to specify each job. I saw the linked mergify question, and the maintainer's response can be interpreted different ways.  I think while the api may _return_ all of these names, you can still specify the top-level name.  I am ok with leaving this for now though, we can always experiment (and check the api, I haven't found exactly where to look yet) and simplify it later.

tox.ini
```diff
@@ -1,12 +1,6 @@
 [tox]
 envlist = py27, py36, py37, flake8
 
-[travis]
```
jguiditta, line 4: Perhaps this is not needed, but when I did this, I changed travis to gh-actions here (could just be a different way and both work, not sure).  See here:
https://github.com/release-depot/flask-container-scaffold/blob/main/tox.ini

jguiditta (commented):
Overall, looks good, couple comments/thoughts.

.github/workflows/test.yml
jpichon: This doesn't seem useful for flake8 since linting isn't version dependent?

.github/workflows/test.yml
jpichon: I did this on purpose indeed, since it seems slower and unnecessary to me to repeat the flake8 step for every version?

tox.ini
jpichon: I tested the job against my branch while working on it and it seems to set up the versions correctly without it: https://github.com/jpichon/git_wrapper/actions/runs/574993890

.mergify.yml
jpichon: I'd love to have this be just one step, it would be clearer, less unwieldy and we wouldn't need to remember to change it with every python version update. I read the maintainer answer as requiring every job name to be exact in the mergify file. I'm not sure about the best way to go about testing this either, suggestions welcome!

tox.ini
jpichon: Then again, that might be why I had to specify -epy: just using "tox" as a command ran zero tests.

Thinking further, I'm tempted to leave it as is because it means updating the version matrix only requires a change in the workflow file, as opposed to remembering to update both test.yml and tox.ini?

.mergify.yml
jpichon: I compressed it into one line - let's be careful with the next MR and see if that works!

.github/workflows/test.yml
jguiditta: Sure, I can see that

.github/workflows/test.yml
jguiditta: Also fair

.mergify.yml
jguiditta: Cool - keep in mind we won't really know until we merge the PR with the mergify change (so meta!).  So it is possible we will need a rev or two if this doesn't work, but at least we know the actions are godd, since we can see how they ran on your branch when you pushed the commit(s)

tox.ini
jguiditta: I can see that as a benefit, and have no issue with it - perhaps we will change the other projects to do the same.

jguiditta (approved):
Actions pass, and as I said in another comment, I think we can rev mergify if it doesn't work as expected.

jpichon:
Looks like the Travis reporting hung again, to highlight how necessary this patch is? ;)

jguiditta:
> Looks like the Travis reporting hung again, to highlight how necessary this patch is? ;)

Of course! I'll manually merge it now, then we can forget travis

jpichon:
Thanks! And now I realised I kept "main" instead of "master" in the branch names so it won't work at all......... I'll push an update.

PR #79, opened 2021-02-19
Fix branch names in github action

jpichon:


jpichon:
Nice to see the github checks running! Although I'm not sure why Travis still shows in here, the configuration shouldn't be on master anymore either. And it says "Required"? I wonder if it's because it's in the webhooks settings? I'm not seeing anything else obvious in the settings.

Interestingly, although the github action tests have all passed, the Mergify configuration doesn't seem to be picking up on them: "status-success=Run tests" still shows as unchecked. So I will use this PR to also adjust the Mergify configuration this round, since the short name unfortunately doesn't seem to work.

jpichon:
Ah, I found the setting, leaving it here in case the problem comes up in a different project since it wasn't obvious to me: when going to the branch settings, there's a list of "protected" branches. I thought that meant "can't delete" but actually when going to edit it there are a bunch of rules and the travis checkbox was checked under this.

fuzzball81 (approved):

jguiditta (approved):

Issue #75, opened 2021-01-11
Add an RST linter in a github action for README.rst and HISTORY.rst


jpichon:
We use a combination of README.rst and HISTORY.rst in the long_description for setup.cfg. That means a parsing error will make the upload to Pypi fail (see for example https://github.com/release-depot/git_wrapper/pull/74).

Adding a github action to check those files ahead of merging would help to prevent this.

jpichon:
This is the command I used locally to check these files, though we might not actually need the output part in an automated check?

    $ python -m readme_renderer README.rst -o /tmp/README.html
    $ python -m readme_renderer HISTORY.rst -o /tmp/HISTORY.html

(cf README instructions at https://github.com/release-depot/git_wrapper#pushing-a-new-release)

jguiditta:
This is just a thought, but I took a look at readme-renderer, and they mention you can run 'twine check' to validate the rst[1].  Would this simplify anything or allow us to not need the extra dep for the renderer?
[1] https://packaging.python.org/guides/making-a-pypi-friendly-readme/#validating-restructuredtext-markup

jpichon:
Gave it a shot and it does seem to pick up the issue as expected, nice!

Current master:
```
$ twine check dist/*
Checking dist/git_wrapper-0.2.8.dev1-py3-none-any.whl: PASSED, with warnings
  warning: `long_description_content_type` missing. defaulting to `text/x-rst`.
Checking dist/git_wrapper-0.2.8.dev1.tar.gz: PASSED, with warnings
  warning: `long_description_content_type` missing. defaulting to `text/x-rst`.
$ echo $?
0
```

After reproducing the old indendation error:

```
$ twine check dist/*
Checking dist/git_wrapper-0.2.8.dev1-py3-none-any.whl: FAILED
  `long_description` has syntax errors in markup and would not be rendered on PyPI.
    line 172: Warning: Block quote ends without a blank line; unexpected unindent.
  warning: `long_description_content_type` missing. defaulting to `text/x-rst`.
Checking dist/git_wrapper-0.2.8.dev1.tar.gz: FAILED
  `long_description` has syntax errors in markup and would not be rendered on PyPI.
    line 172: Warning: Block quote ends without a blank line; unexpected unindent.
  warning: `long_description_content_type` missing. defaulting to `text/x-rst`.
$ echo $?
1
```

So I agree that we should update the README, as it's much simpler to use the same tool for both checking and uploading.

[...]

PR #82, opened 2021-06-22
Create GitHub Action CI job for twine check

kyleecodes:
To prevent upload failures to PyPi due to linting
errors in .rst files, a CI job for running twine
check was made into a GitHub Action

twine check effectively removes the need for an
additional dep for linting .rst files

The variable long_description_content_type was
added to setup.cfg in order to pass twine check
--strict

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,3 +51,17 @@ jobs:
             pip install tox tox-gh-actions
         - name: Run integration tests
           run: tox -eintegration
+    twine:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v2
+        - name: Setup Python
+          uses: actions/setup-python@v2
+          with:
+            python-version: 3.9
+        - name: Install Tox and any other packages
+          run: |
+            python -m pip install --upgrade pip
+            pip install tox tox-gh-actions
+        - name: Run twine-check
+          run: tox -etwine
--- a/README.rst
+++ b/README.rst
@@ -74,14 +74,9 @@ Pushing a new release
 
    You may want to check the HISTORY.rst file renders correctly before
    committing, as the twine upload will fail later if it
-   doesn't. Install one of these::
+   doesn't::
 
-    # dnf install python3-readme-renderer
-    $ pip install readme-renderer
-
-   Then run the following command::
-
-    $ python -m readme_renderer HISTORY.rst -o /tmp/HISTORY.html
+    $ tox -etwine
 
 2. Once that's merged, tag that patch and push the new tag to the repo
 3. Run the following commands::
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -4,3 +4,4 @@ pytest-datadir
 pytest-runner
 mock>=2.0.0
 flake8
+twine
--- a/tox.ini
+++ b/tox.ini
@@ -26,3 +26,8 @@ commands =
     /usr/bin/find . -name '*.pyc' -delete  # Avoid issues with pytests conf file discovery
     /usr/bin/podman build -t git_wrapper_integration_tests .
     /usr/bin/podman run git_wrapper_integration_tests /bin/sh -c "pytest -k integration_tests integration_tests"  # Ensure we only run the integration tests
+
+[testenv:twine]
+commands =
+    python setup.py sdist bdist_wheel
+    twine check --strict dist/*
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,7 @@ ENV/
 # Pipenv
 Pipfile
 Pipfile.lock
+
+# auto-generated
+ChangeLog
+AUTHORS
--- a/setup.py
+++ b/setup.py
@@ -7,5 +7,6 @@ from setuptools import setup
 
 setup(
     setup_requires=['pbr'],
-    pbr=True
+    pbr=True,
+    long_description_content_type='text/x-rst'
 )


AUTHORS
```diff
@@ -0,0 +1,7 @@
+Cédric Jeanneret <cjeanner@redhat.com>
+Haikel Guemar <hguemar@fedoraproject.org>
+Jason Guiditta <jason.guiditta@gmail.com>
+Jason Joyce <fuzzball81@gmail.com>
+Jason Joyce <jjoyce@redhat.com>
+Julie Pichon <jpichon@redhat.com>
+kyleecodes <kfieldsinmd@gmail.com>
```
jpichon, line 7: I believe this was automatically generated, we don't want to commit this.

ChangeLog
```diff
@@ -0,0 +1,134 @@
+CHANGES
```
jpichon, line 1: This duplicates HISTORY.rst, I don't think this file should have been added.

Maybe these two files could be added to .gitignore instead to avoid that kind of confusion in the future.

README.rst
```diff
@@ -74,14 +74,10 @@ Pushing a new release
...
+   doesn't::
 
-    # dnf install python3-readme-renderer
-    $ pip install readme-renderer
-
-   Then run the following command::
-
-    $ python -m readme_renderer HISTORY.rst -o /tmp/HISTORY.html
+    $ python setup.py sdist bdist_wheel
+    $ twine check dist/*
```
jpichon, line 81: Thank you for updating the README file!

.github/workflows/test.yml
```diff
@@ -51,3 +51,17 @@ jobs:
...
+        - name: Setup Python
+          uses: actions/setup-python@v2
+          with:
+            python-version: 3.9
+        - name: Install Tox and any other packages
+          run: |
+            python -m pip install --upgrade pip
+            pip install tox tox-gh-actions
+        - name: Run twine-check
+          run: tox -etwine
```
jpichon, line 67: It looks like the little warning sign here means there's no newline at the end of the file, you may want to add them to the affected files.

AUTHORS
```diff
@@ -0,0 +1,7 @@
+Cédric Jeanneret <cjeanner@redhat.com>
+Haikel Guemar <hguemar@fedoraproject.org>
+Jason Guiditta <jason.guiditta@gmail.com>
+Jason Joyce <fuzzball81@gmail.com>
+Jason Joyce <jjoyce@redhat.com>
```
jguiditta, line 5: I am not sure we want both of Jason's emails here? Don't feel strongly, just seems kind of pointless to have both.

.github/workflows/test.yml
jguiditta, line 67: This line shows a warning of 'no newline', should we add one to the end of the file?

tox.ini
```diff
@@ -26,3 +26,8 @@ commands =
     /usr/bin/find . -name '*.pyc' -delete  # Avoid issues with pytests conf file discovery
     /usr/bin/podman build -t git_wrapper_integration_tests .
     /usr/bin/podman run git_wrapper_integration_tests /bin/sh -c "pytest -k integration_tests integration_tests"  # Ensure we only run the integration tests
+
+[testenv:twine]
+commands =
+    python setup.py sdist bdist_wheel
+    twine check --strict dist/*
```
jguiditta, line 33: Same warning about newline here.

jpichon, line 33: (Nit) Same newline tweak here.

jguiditta (commented):
Overall, looks great, actions pass! Couple questions inline.  Also, not a big deal. but this probably should be separate PRs - one to add the github action, and a separate one adding things like AUTHORS and Changelog, as they don't seem related (unless they were causing the job to fail if they were not there).

AUTHORS
kyleecodes: Must be the result of an automatic process, wasn't sure if I should mess with it. There is already the AUTHORS.rst.

jpichon (changes requested):
Looking good, and the job results seem fine as well! Some automatically generated files were included in the commit that should be removed, as well as a couple of other nits. Could you also add a reference to "Issue #75" in the commit? Github will link it automatically this way - https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls#issues-and-pull-requests

.github/workflows/test.yml
kyleecodes: Definitely!

AUTHORS
kyleecodes: Correct it is, will remove it with updated patch

ChangeLog
kyleecodes: Good idea! Best to prevent this for the future.

jguiditta (dismissed):
I think this looks good now!

test-requirements.txt
```diff
@@ -4,3 +4,5 @@ pytest-datadir
 pytest-runner
 mock>=2.0.0
 flake8
+twine
+wheel
```
fuzzball81, line 8: Do we need wheel here?

README.rst
```diff
@@ -74,14 +74,10 @@ Pushing a new release
...
```
fuzzball81, line 80: I think we want this to match what was added to tox.

fuzzball81 (commented):
Overall looks good, a few questions in line.

test-requirements.txt
kyleecodes: tox references this file for dependencies, with flake8 also there, I added it preemptively. On my machine however, tox will run the check without these here. I can remove these and as long as GH actions doesn't need them, then they aren't needed at all!

kyleecodes: Tested this on a test branch, twine is required to pass through GH Actions, but wheel is not. So I will remove wheel.

Used a test branch to avoid branch naming inconsistencies, GH actions requires the name of the branch in order to run the workflow, and my branch and master have two different names.

README.rst
jpichon: Maybe it makes sense to use the tox job directly, even?

test-requirements.txt
jpichon: Thanks for testing, sounds good to remove what's not actually necessary.

README.rst
kyleecodes: Yes using the tox command would get rid of the need for the sdist command too.

jpichon (approved):
LGTM!

jpichon:
Mergify message reads "GitHub App like Mergify are not allowed to merge pull request where .github/workflows is changed.", merging manually.

Issue #32, opened 2018-09-06
Drop Python 2 compatibility library


jpichon:
Python-future was added in PR #31 in order to temporarily add support for Python 2 but should be removed as soon as reasonable. (The dependency on wrapt can quite probably be removed at the same time as well, if we want to.)

PR #83, opened 2021-07-21
Modify to allow tag listing

kyleecodes:
Adds the method list_tags() to the GitTag class to list all tags in a repo object.

### DIFF
--- a/git_wrapper/tag.py
+++ b/git_wrapper/tag.py
@@ -72,3 +72,14 @@ class GitTag(object):
                 self.git_repo.git.push(remote, name)
         except git.GitCommandError as ex:
             raise_from(exceptions.PushException(msg), ex)
+
+    def names(self):
+        """List git tags in the repository."""
+        try:
+            tags_list = [x.name for x in self.git_repo.repo.tags]
+        except git.GitCommandError as ex:
+            repo_path = self.git_repo.repo.working_dir
+            msg = "Error listing tags for {repo}. Error: {error}".format(error=ex, repo=repo_path)
+            raise_from(exceptions.TaggingException(msg), ex)
+
+        return tags_list
--- a/tests/test_tag.py
+++ b/tests/test_tag.py
@@ -1,7 +1,7 @@
 #! /usr/bin/env python
 """Tests for GitTag"""
 
-from mock import patch
+from mock import patch, PropertyMock
 
 import git
 import pytest
@@ -155,3 +155,41 @@ def test_push_tag_failed(mock_repo):
     with patch('git.repo.fun.name_to_object'):
         with pytest.raises(exceptions.PushException):
             repo.tag.push("my_tag", "origin")
+
+
+def test_names(mock_repo):
+    """
+    GIVEN GitRepo is initialized with a path and repo
+    WHEN GitTag.names() is called
+    THEN repo.tags is returned
+    """
+    class TestTag(object):
+
+        def __init__(self, name):
+            self.name = name
+
+    tag1 = TestTag('tag1')
+    tag2 = TestTag('tag2')
+    expected = [tag1, tag2]
+    expected_names = [x.name for x in expected]
+
+    type(mock_repo).tags = PropertyMock(return_value=expected)
+
+    repo = GitRepo(repo=mock_repo)
+    test_tags = repo.tag.names()
+
+    assert expected_names == test_tags
+
+
+def test_names_failed(mock_repo):
+    """
+    GIVEN GitRepo is initialized with a path and repo
+    WHEN GitTag.names() is called
+    AND repo.tags fails
+    THEN a TaggingException is raised
+    """
+    type(mock_repo).tags = PropertyMock(side_effect=git.GitCommandError('tags', ''))
+    repo = GitRepo(repo=mock_repo)
+
+    with pytest.raises(exceptions.TaggingException):
+        repo.tag.names()
--- a/integration_tests/test_tag.py
+++ b/integration_tests/test_tag.py
@@ -11,6 +11,9 @@ def test_tag(repo_root):
     repo.git.checkout("master")
     head = repo.repo.head.object.hexsha
 
+    # Test tag listing function names() by checking for tag name in test repo
+    assert '0.2.4' in repo.tag.names()
+
     # Test tag creation by tagging the current head
     repo.tag.create("test_tag", head)
 


tests/test_tag.py
```diff
@@ -155,3 +156,37 @@ def test_push_tag_failed(mock_repo):
...
+
+def test_list_tags(mock_repo):
+    """
+   GIVEN GitRepo is initialized with a path and repo
+   WHEN GitTag.list_tags() is called
+   THEN repo.tags is returned
+   """
+    expected = ['tag1', 'tag2']
+    my_local_mock = mock_repo
+    type(my_local_mock).tags = PropertyMock(return_value=expected)
```
jpichon, line 169: This looks off to me, since type() is a function that returns a type object, and we're trying to assign a value to that? Maybe this will be easier to do once list() returns strings. The unit test for mocking remote names may also be of interest, as another way to go about mocking objects.

tests/test_tag.py
```diff
@@ -155,3 +156,37 @@ def test_push_tag_failed(mock_repo):
...
+    repo = GitRepo(repo=my_local_mock)
+
+    test_tags = GitTag(repo, None).list_tags()
```
jpichon, line 172: We should use repo.tag like in the other tests, rather than GitTag(), to stay close to the way the library should actually be used.

tests/test_tag.py
```diff
@@ -155,3 +156,37 @@ def test_push_tag_failed(mock_repo):
...
+   GIVEN GitRepo is initialized with a path and repo
+   WHEN GitTag.list_tags() is called
+   AND repo.tags fails
+   THEN a TaggingException is raised
+   """
+    def func():
+        raise git.GitCommandError('tags', '')
+
+    my_local_mock = mock_repo
+    type(my_local_mock).tags = PropertyMock(side_effect=func)
```
jpichon, line 188: I'm just as suspicious about type() here, if the tests and mocking are still causing issues I can try and have a look once the code is updated, but this doesn't look right to me.

tests/test_tag.py
```diff
@@ -155,3 +156,37 @@ def test_push_tag_failed(mock_repo):
...
+    repo = GitRepo(repo=mock_repo)
+
+    with pytest.raises(exceptions.TaggingException):
+        GitTag(repo, None).list_tags()
```
jpichon, line 192: Same note about calling repo.tag.list_tags()

git_wrapper/tag.py
```diff
@@ -72,3 +72,16 @@ def push(self, name, remote, dry_run=False):
...
+
+    def list_tags(self):
+        """List git tags in the repository.
+        """
+
+        try:
+            tags_list = self.git_repo.repo.tags
+        except git.GitCommandError as ex:
+            tags_list = []
+            msg = "Error listing tags for {repo}. Error: {error}".format(error=ex, repo=self.git_repo)
```
jpichon, line 84: self.git_repo is an object so it doesn't work well as an error message, you end up with output like this:

git_wrapper.exceptions.TaggingException: Error listing tags for <git_wrapper.repo.GitRepo object at 0x7f11703b9fd0>. Error: 

I'm not sure if it's absolutely necessary to have the repo name in this, since the caller should know what they initialised GitRepo with, but if it's there, it should be a human-readable string.

jpichon (changes requested):
Please don't open PRs unless they are actually ready to review. The [DO NOT MERGE] marker is gone so I'm assume it's ready now, but it's really confusing for reviewers.

tests/test_tag.py
fuzzball81: PropertyMocks can't be attached directly to a mock object like mock_repo, instead they need to be attached to they type, please see https://docs.python.org/3/library/unittest.mock.html#unittest.mock.PropertyMock for more information.

I agree it looks odd, however this does allow of the control over what is expected. You could modify the conftest.py to add a 'tags' mock but would not have control over what was returned without also adding additional arguments to the fixture.

tests/test_tag.py
fuzzball81: See the above comment on type usage with PropertyMock.

tests/test_tag.py
jpichon: Right! I think what we did for the remote names with mock.configure() would be an option as well, but there are advantages to this way as well, it's a lot more concise. Thanks for the link!

jguiditta:
> Please don't open PRs unless they are actually ready to review. The [DO NOT MERGE] marker is gone so I'm assume it's ready now, but it's really confusing for reviewers.

That is my fault. Kylee was having trouble getting her tests to work, so I asked her to put it up this way so we could help her figure out the problem.

git_wrapper/tag.py
```diff
@@ -72,3 +72,19 @@ def push(self, name, remote, dry_run=False):
                 self.git_repo.git.push(remote, name)
         except git.GitCommandError as ex:
             raise_from(exceptions.PushException(msg), ex)
+
+
+    def list_tags(self):
```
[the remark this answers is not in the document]
kyleecodes: True, list() is less redundant. But my logic behind list_tags() was to stay away from using names of Pythons builtin functions. https://docs.python.org/3/library/functions.html#func-list.
Naming user-defined functions the same name as built-in functions can confuse interpreters, especially in popular IDEs like VSCode.

tests/test_tag.py
```diff
@@ -155,3 +155,44 @@ def test_push_tag_failed(mock_repo):
...
             repo.tag.push("my_tag", "origin")
+
+
+def test_list_tags(mock_repo):
+    """
+    GIVEN GitRepo is initialized with a path and repo
+    WHEN GitTag.list_tags() is called
+    THEN repo.tags is returned
+    """
+    class test_tag(object):
```
jpichon, line 166: In general, class names should use CapitalisedWords as per pep8 conventions https://www.python.org/dev/peps/pep-0008/#class-names

tests/test_tag.py
```diff
@@ -155,3 +155,44 @@ def test_push_tag_failed(mock_repo):
...
+    """
+   GIVEN GitRepo is initialized with a path and repo
+   WHEN GitTag.list_tags() is called
+   AND repo.tags fails
+   THEN a TaggingException is raised
+   """
+    def func():
+        raise git.GitCommandError('tags', '')
+
+    type(mock_repo).tags = PropertyMock(side_effect=func)
```
jpichon, line 194: I think the extra function is unnecessary, `side_effect=git.GitCommandError('tags', '')` gives similar results in my testing.

git_wrapper/tag.py
```diff
@@ -72,3 +72,15 @@ def push(self, name, remote, dry_run=False):
                 self.git_repo.git.push(remote, name)
         except git.GitCommandError as ex:
             raise_from(exceptions.PushException(msg), ex)
+
+    def list_tags(self):
+        """List git tags in the repository.
+        """
```
jpichon, line 78: (Nit) For documentation strings that fit on a single line, the pep8 recommendation is to place them on the same line (cf. https://www.python.org/dev/peps/pep-0008/#documentation-strings )

git_wrapper/tag.py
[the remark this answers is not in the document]
jpichon: Maybe "names()" like for the remote listing would be an option then, and consistent with the other code we have? Just a suggestion, I won't block on this.

jpichon (commented):
This is looking a lot better!! Thank you for the update, I just made a couple of minor comments inline.

jpichon (commented):
LGTM!

jpichon (approved):

git_wrapper/tag.py
```diff
@@ -72,3 +72,14 @@ def push(self, name, remote, dry_run=False):
...
         except git.GitCommandError as ex:
             raise_from(exceptions.PushException(msg), ex)
+
+    def names(self):
+        """List git tags in the repository."""
+        try:
+            tags_list = [x.name for x in self.git_repo.repo.tags]
+        except git.GitCommandError as ex:
+            repo_path = self.git_repo.repo.working_dir
+            msg = "Error listing tags for {repo}. Error: {error}".format(error=ex, repo=repo_path)
```
jguiditta, line 82: Minor, but maybe in next patch, we can switch this to use fstrings?

jguiditta (commented):
I think this looks fine as well (post-merge review, just a sanity check).  One minor request for later, but not a big deal.

git_wrapper/tag.py
jpichon: The library is still nominally Python 2-compatible, I think there are a lot of places that need to be updated. It probably should be done all at once, feel free to add notes to https://github.com/release-depot/git_wrapper/issues/32 in relation to this work. And thank you for sanity checking the patch!

PR #85, opened 2021-07-27
Add publish action.

jguiditta:
Remove some manual steps/effort by having new releases be published
automatically to pypi.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,28 @@
+name: Upload Package to Pypi
+
+on:
+  release:
+  workflow_dispatch:
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r test-requirements.txt
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        python setup.py sdist bdist_wheel
+        twine upload dist/*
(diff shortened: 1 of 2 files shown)


jpichon:
LGTM! Since the twine check is a job now, it seems fine to remove the manual step related to it. (Although I just notice now this job isn't showing as 'required' in the list of checks, we might want to fix that as well.) Really nice simplification of the workflow.

fuzzball81 (dismissed):

jguiditta:
> LGTM! Since the twine check is a job now, it seems fine to remove the manual step related to it. (Although I just notice now this job isn't showing as 'required' in the list of checks, we might want to fix that as well.) Really nice simplification of the workflow.

Well, this job is not required, and should only run on release or 'workflow_dispatch' (which s just a way to allow you to manually trigger the job).  I set this up the same way we have it in flask-container-scaffold, so I think it should be fine.  If you know of a specific change you think we should make, I am happy to do so though.

jpichon:
I was talking about the twine check we recently added to make sure the history and readme files are correctly formatted. I do think it should be required, because if that job fails we won't be able to release until whatever file is fixed. But this isn't directly related to this patch/PR either, so it's fine, more of a note. This PR might just need a rebase, then it is good to go.

jguiditta:
@jpichon ah, ok, I misunderstood what you meant.  I agree we want twine check to run first.  My first thought though, is that this is a different flow altogether.  Twine check is run on every commit, so we should never have the merged codebase in a state where that fails, unless I have missed something.  This job will ONLY run when a tag (and thus a release) is pushed.  I guess the only case where we could accidentally blow something up is if someone manually kicked off a build with this, though I am still not sure how we would have merged a commit that broken the check.

That aside, if there is a reasonable way to import and depend on a job from a different file, I have no issue with that.  Or, we could simply add the twine check here as well (this is actually where I do it in the other project).  It is a single line, and runs quickly.  That change I could even make to this PR if you wanted.

jpichon (approved):

fuzzball81 (approved):

jpichon:
PRs touching workflow files must be merged manually, doing that now.

PR #86, opened 2021-07-29
Update outdated development/contrib instructions

jpichon:
Pipenv was removed in an early commit (bb5528) but the documentation wasn't updated to reflect that.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -64,11 +64,15 @@ Ready to contribute? Here's how to set up `git_wrapper` for local development.
 
     $ git clone git@github.com:your_name_here/git_wrapper.git
 
-3. Use pipenv via the Makefile to setup a new virtualenv. Pipenv will create a new virtualenv with packages at the correct versions.
-   Assuming you have pipenv installed, this is how you set up your fork for local development::
+3. Use the Makefile or virtualenv to setup a new virtualenv::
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -64,11 +64,15 @@ Ready to contribute? Here's how to set up `git_wrapper` for local development.
...
-   Assuming you have pipenv installed, this is how you set up your fork for local development::
+3. Use the Makefile or virtualenv to setup a new virtualenv::
 
     $ make dev
-    $ pipenv shell
+
+   or::
+
+    $ virtualenv .venv
+    $ source .venv/bin/activate
```
fuzzball81, line 74: If your using pipenv I don't think you need to run these steps first. `pipenv install -r requirements.txt` should create the venv as part of it's setup.

jpichon: We removed a bunch of pipenv references in https://github.com/release-depot/git_wrapper/commit/bb552842c21b2a5dcf72879bafad7b0788106b86 and it doesn't seem to work well when I try it now, so I removed the last references to it with this patch. I know I personally never use it when working on git_wrapper, but we can leave the refs if someone wants to figure out and fix up the broken bits.

jpichon:
Superseded by https://github.com/release-depot/git_wrapper/pull/89/

PR #89, opened 2021-08-05
Revamp our build process to be more in line with our other projects.

jguiditta:
We are trying to stick with following the most modern build process,
as recommended by pypa. Also, this changes the process to use
setuptools_scm, which gets the version information from git.


jguiditta:
I imported this as a test repo in RTD and it builds as expected.  If you want to see what the docs look like without building them yourself, you can view my (temporary) docs here:
https://test-git-wrapper.readthedocs.io/en/latest/

Makefile
```diff
@@ -52,47 +52,40 @@ clean-test: ## remove test and coverage artifacts
 lint: ## check style with flake8
 	pipenv run flake8 --ignore=E501 git_wrapper tests integration_tests
 
-test: dev ## run tests quickly with the default Python
-	pipenv run py.test tests
+test: ## run tests quickly with the default Python
```
fuzzball81, line 55: Do we need the Makefile at this point? If it's still providing value then we can keep it. I have a feeling very feel people will actually use it, especially if we have automation around building.

fuzzball81 (commented):
Looks good overall. Nice work!

Makefile
```diff
@@ -52,47 +52,40 @@ clean-test: ## remove test and coverage artifacts
...
+	pipenv run $(MAKE) -C docs html
 	$(BROWSER) docs/_build/html/index.html
 
 servedocs: docs ## compile the docs watching for changes
 	pipenv run watchmedo shell-command -p '*.rst' -c '$(MAKE) -C docs html' -R -D .
 
 release: clean ## package and upload a release
-	pipenv run python setup.py sdist upload
-	pipenv run python setup.py bdist_wheel upload
+	pipenv run twine check dist/*
```
jpichon, line 74: We seem to check --strict in the tox job used in the gh action, should we do that here too to stay consistent and avoid surprises? I wonder if we should call the Make commands where possible in the gh actions too if we're keeping the Makefile, to avoid more duplication.

git_wrapper/__init__.py
```diff
@@ -4,4 +4,3 @@
 
 __author__ = """Jason Joyce"""
 __email__ = 'fuzzball81@gmail.com'
-__version__ = '0.2.8'
```
jpichon, line 7: Oh, is that no longer necessary? We may want to update the README file section on releasing to indicate that only the HISTORY file needs to be updated, then?

I wonder if it would be worth encouraging new features to come with a HISTORY update from now on too, so we'll only need to tag to prepare a new release rather than having to submit a patch first? If I understand correctly. Maybe writing up the highlights all at once is fine.

jpichon (commented):
Really nice clean-ups!!

git_wrapper/__init__.py
jguiditta: The plan here is to use setuptools_scm, which automatically generates the version based on the git tag.  So, when we do a release, the publish job should have a version of the project checked out by tag for newest release and build from that.  I tested this locally, and it seemed to work as expected.  If you do a build without a tag, it will build with <most-recent-tag>.dev<#>+<git hash> so you can install a dev build easily.  
For now, yes, only the HISTORY file would need to be updated for a release, after tagging.  I would like to make this more automated as well, though the other option would be to do a more curated changelog that describes features and bugs fixed, rather than simply listing commits between releases.  It is possible we may even want both, I am not really sure yet.  Either way, I propose we consider this a future enhancement, otherwise I am afraid I'll never finish this patch.  And yes, I'll update the relevant section of the readme in an update shortly.

Makefile
jguiditta: Great idea, I'll change this to call tox -etwine like the github action.

Makefile
jguiditta: I know I originally suggested we simply toss the Makefile, and yes, we still could do that if everyone wants.  However, while working on this patch, and trying to update the Makefile, I actually found it super handy, especially for being able to run 'make docs', and have it clean, build and launch in a browser, vs having to 'cd docs/;make clean; make html; <navigate to file in browser'.  Perhaps I should add the alternative direct pipenv (or even virtualenv, like in @jpichon 's other patch, which I hadn't even seen) directions to the README or CONTRIBUTING docs.

jguiditta: Also, I think we should switch to having the make/virtualenv/pipenv directions only in one place, I didn't notice before they were duplicated.

README.rst
```diff
@@ -57,30 +48,31 @@ To build the documentation on your checkout, simply run::
...
-We plan to get this published in the near future, and this README will be
-updated when that happens.
-
 Contributions
 *************
 
 All new code should include tests that excercise the code and prove that it
 works, or fixes the bug you are trying to fix.  Any Pull Request without tests
-will not be accepted.
+will not be accepted. See `CONTRIBUTING <CONTRIBUTING.rst>`_ for more details.
```
jguiditta, line 56: I notice this works here in github, but not on the sphinx-generated doc.  I can update it to be just text, if that is what everyone prefers, just let me know.

jpichon: Interesting! Personally, I think CONTRIBUTING.rst without link might be just as well in this case. Easier for RTD readers to understand they need to look for a file.

Makefile
jpichon: I wonder if we might just remove the parts we don't use and keep e.g. the docs part only, but it seems fine to punt that conversation, especially now that many of the references are to tox jobs and thus avoid some of the duplication.

README.rst
```diff
@@ -57,30 +48,31 @@ To build the documentation on your checkout, simply run::
...
 
 .. _example: https://github.com/release-depot/git_wrapper/commit/fc88bcb3158187ba9566dad896e3c688d8bc5109
-
-Credits
--------
-
-This package was created with Cookiecutter_ and the `audreyr/cookiecutter-pypackage`_ project template.
-
-.. _Cookiecutter: https://github.com/audreyr/cookiecutter
-.. _`audreyr/cookiecutter-pypackage`: https://github.com/audreyr/cookiecutter-pypackage
```
jpichon, line 86: We might want to just say "based on but heavily modified"?

jpichon (commented):
Nit to answer the question about the CONTRIBUTING.rst file, otherwise LGTM

README.rst
jguiditta: K, updating

README.rst
jguiditta: I suppose we could keep this, but alter the description.  My thought was that this is so far from how it was initially set up years ago, that what we used to create it at that time was no longer relevant.  If we want to keep it for historical reference, I am ok with that.

Makefile
jguiditta: Yeah, personally, I would like to keep it, at least for now.  I know I am not a primary contributor to this codebase, but just in doing these packaging changes, I used it extensively to test setting up a fresh environment, running tests, building docs, etc.  Doing it that way actually exposed a number of errors I was able to correct in my changes.  There is also a convenience level to running `make tests` vs `pipenv run tox -epy`, though I know that is more subjective.

jpichon: This is fine by me if it's proven useful! And it might be now that it's up to date again.

README.rst
jguiditta: Ok, so I changed the contributing link to just be test, I am unsure if we have agreed what to do with this part.  If you want it in, I'll add it, otherwise, I think I have addressed the remaining concerns.

jpichon (approved):
LGTM!

fuzzball81 (approved):

Issue #88, opened 2021-07-29
Fix "Upload Package to Pipy" workflow


jpichon:
The "Upload package to Pypi" workflow doesn't work.

The job fails with the following traceback:

 ```
 python setup.py sdist bdist_wheel
  twine upload dist/*
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.9.6/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.9.6/x64/lib
    TWINE_USERNAME: ***
    TWINE_PASSWORD: ***
/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/setuptools/dist.py:642: UserWarning: Usage of dash-separated 'author-email' will not be supported in future versions. Please use the underscore name 'author_email' instead
  warnings.warn(
/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/setuptools/dist.py:642: UserWarning: Usage of dash-separated 'description-file' will not be supported in future versions. Please use the underscore name 'description_file' instead
  warnings.warn(
/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/setuptools/dist.py:642: UserWarning: Usage of dash-separated 'home-page' will not be supported in future versions. Please use the underscore name 'home_page' instead
  warnings.warn(
WARNING: The wheel package is not available.
/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/setuptools/dist.py:642: UserWarning: Usage of dash-separated 'author-email' will not be supported in future versions. Please use the underscore name 'author_email' instead
  warnings.warn(
/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/setuptools/dist.py:642: UserWarning: Usage of dash-separated 'description-file' will not be supported in future versions. Please use the underscore name 'description_file' instead
  warnings.warn(
/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/setuptools/dist.py:642: UserWarning: Usage of dash-separated 'home-page' will not be supported in future versions. Please use the underscore name 'home_page' instead
  warnings.warn(
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
[...]

PR #91, opened 2021-08-31
Improvements to publish github action.

jguiditta:
This patch attempts to any remaining issues from #88, specifically
the mutliple runs (by specifying one one python version to build for)
and not requiring a 'click release button' step to be added to our
process. We may, in a future patch, decide to do more around the
github 'release' concept, which is separate from our current process
of doing a release to pypi when the project is tagged.

This patch also adds a push to the test instance of pypi for quicker
and more frequent verification that pulishing works as expected, and
to give us an easy way for new features to be tested by users between
official tags/releases.

This patch uses the github action built by pypa instead of twine
directly, as we did before, following official docs [1],[2].

[1]
https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
[2] https://github.com/pypa/gh-action-pypi-publish

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,12 +1,11 @@
-name: Upload Package to Pypi
+name: Upload Package to PyPI and TestPyPI
 
 on:
-  release:
+  push:
   workflow_dispatch:
 
 jobs:
   deploy:
-
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -1,28 +1,36 @@
...
+    - name: Publish to Test PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.TEST_PYPI_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+    - name: Publish to PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.TEST_PYPI_TOKEN }}
```
fuzzball81, line 32: Does this need to be a different token, or is the same used for test and release?

fuzzball81 (commented):
Overall this looks good. One minor question.

.github/workflows/publish.yml
jguiditta: These are different tokens (and I already created them and added them to github)

fuzzball81: Let me rephrase. Should we be using TEST_PYPI_TOKEN here or a non-test token?

jguiditta: Ah! the snippet here doesn't show enough of the code, so I didn't realize what you meant.  Good catch, this line should be using the non-test one, will update

fuzzball81 (approved):
LGTM

.github/workflows/publish.yml
```diff
@@ -1,28 +1,34 @@
...
+      run:  python -m build
+    - name: Publish to Test PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.TEST_PYPI_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+    - name: Publish to PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
jpichon, line 31: Ooh I was wondering when I saw "push" if we wouldn't be releasing too often. I like this. We're unlikely to tag anything other than releases so this makes sense to me.

jpichon (approved):

jpichon:
Merging manually since changes to github actions cannot be merged automatically.

PR #92, opened 2021-09-02
Fix deploy job.

jguiditta:
We were installing the test requirements instead of dist, so the build
would fail.


jguiditta:
Action trigger on my fork fails on missing token instead of build, as expected:
https://github.com/jguiditta/git_wrapper/runs/3494679320?check_suite_focus=true

jpichon (approved):
LGTM, thanks for the quick fix!

fuzzball81 (approved):
LGTM

PR #93, opened 2021-09-03
Change how we generate version for test pypi builds.

jguiditta:
The tooling by default appends '+{hash}' to non-tagged builds,
which is not compliant with the format required by pypi. This
patch chops off the '+' bit, so you get a version like '0.2.9.dev5'
where 'dev5' is the number of commits from the last tag. Provided
github action puts the env var where the python tooling expects it,
this should solve the issue for test deploys (production is unchanged).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,9 +18,17 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r dist-requirements.txt
+    - name: Create Environment Variable For Test PyPI Builds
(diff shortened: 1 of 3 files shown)


jguiditta:
since the new job wont run until merge, here are the results from pushing commit to my fork:
https://github.com/jguiditta/git_wrapper/runs/3507023658?check_suite_focus=true
Note that the fail at 'push to test pypi' step is expected because I do not have an api token set up on my repo.

jpichon (approved):
LGTM

fuzzball81 (approved):
LGTM

PR #94, opened 2021-09-07
Change fetch-depth to get correct tag info for builds.

jguiditta:
The default setting of the checkout@v2 action sets this to
1, meaning it gets only the single commit and NO data that
setuptools_scm can use to determine what version to build.
According to the docs[1], setting this to 0 should return
the data we need.  Note that this should ONLY be needed for
the publish action, as the tests and twine check don't need
to care about the version data.

[1]
https://github.com/marketplace/actions/checkout#fetch-all-history-for-all-tags-and-branches

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,6 +10,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Set up Python
       uses: actions/setup-python@v2
       with:


jguiditta:
This seems to work as expected on my fork, you can see we are no longer grabbing 0.1 for the build version:

https://github.com/jguiditta/git_wrapper/runs/3534693220?check_suite_focus=true#step:8:41

fuzzball81 (approved):

lhh (approved):

jpichon (approved):
LGTM. Merging manually since it touches workflow files.

PR #96, opened 2021-09-09
Remove python 2-specific code.

jguiditta:
Since we no longer support python 2, there is no need to use
things only needed for compatibility anymore.


jpichon (approved):
LGTM. I'll update the github project config to stop looking for 2.7 also.

fuzzball81 (approved):

jpichon:
Manually merging, since the mergify configuration also changed.

PR #97, opened 2021-09-15
Switch to f-string over .format method.

jguiditta:
The f-string format is more stylistically python 3 than the format
method, and subjectively, easier to read.  This patch switches all
uses of .format to use f-strings to provide a consistent base
going forward.

### DIFF
--- a/git_wrapper/branch.py
+++ b/git_wrapper/branch.py
@@ -28,7 +28,8 @@ class GitBranch(object):
         """
         full_path = os.path.realpath(os.path.expanduser(path))
         if not os.path.isfile(full_path):
-            raise exceptions.FileDoesntExistException('{path} is not a file.'.format(path=full_path))
+            msg = f"{full_path} is not a file."
+            raise exceptions.FileDoesntExistException(msg)
         return full_path
 
     def _run_cherry(self, upstream, head, regex):
@@ -88,7 +89,7 @@ class GitBranch(object):
         else:
             if remote not in self.git_repo.remote.names():
                 raise exceptions.RemoteException(
-                    "Remote {0} does not exist.".format(remote)
+                    f"Remote {remote} does not exist."
                 )
             if name in self.git_repo.repo.remotes[remote].refs:
                 return True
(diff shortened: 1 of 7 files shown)


git_wrapper/branch.py
```diff
@@ -182,7 +185,8 @@ def apply_patch(self, branch_name, path, keep_square_brackets=False):
             else:
                 self.git_repo.git.am(full_path)
         except git.GitCommandError as ex:
-            msg = "Could not apply patch {path} on branch {name}. Error: {error}".format(path=full_path, name=branch_name, error=ex)
+            msg = (f"Could not apply patch {full_path} on branch {branch_name}."
```
jpichon, line 188: Would need a space after the period, I think.

git_wrapper/branch.py
```diff
@@ -129,22 +130,24 @@ def rebase_to_hash(self, branch_name, hash_):
...
-            msg = "Could not checkout branch {name}. Error: {error}".format(name=branch_name, error=ex)
+            msg = f"Could not checkout branch {branch_name}. Error: {ex}"
             raise exceptions.CheckoutException(msg) from ex
 
         # Rebase
         try:
             self.git_repo.git.rebase(hash_)
         except git.GitCommandError as ex:
-            msg = "Could not rebase hash {hash_} onto branch {name}. Error: {error}".format(hash_=hash_, name=branch_name, error=ex)
+            msg = (f"Could not rebase hash {hash_} onto branch {branch_name}."
```
jpichon, line 149: Missing a space after the period.

jpichon (commented):
Couple of minor comments, lgtm otherwise.

git_wrapper/branch.py
jguiditta: Those are good catches, thanks! While I was fixing those, I noticed a couple of our multi-line messages were not consistent regarding whether they broke to a new line for parens.  Since we seemed to have more that did NOT break, I went with that.  If you strongly prefer the other formatting of the parens, I can switch things that way instead, just let me know.

git_wrapper/branch.py
```diff
@@ -228,7 +234,7 @@ def abort_patch_apply(self):
         try:
             self.git_repo.git.am('--abort')
         except git.GitCommandError as ex:
-            msg = "Failed to abort git am operation. Error: {0}".format(ex)
+            msg = "Failed to abort git am operation. Error: {ex}"
```
jpichon, line 237: Ah, just spotted a missing 'f' here - sorry for missing it the first time around.

git_wrapper/branch.py
```diff
@@ -238,12 +244,13 @@ def reverse_diff(self, diff_path):
...
         if not os.path.isfile(full_path):
-            raise exceptions.FileDoesntExistException('{path} is not a file.'.format(path=full_path))
+            msg = f"{full_path} is not a file."
+            raise exceptions.FileDoesntExistException(msg)
 
         try:
             self.git_repo.git.apply(full_path, reverse=True)
         except git.GitCommandError as ex:
-            msg = "Reversing diff failed. Error: {0}".format(ex)
+            msg = "Reversing diff failed. Error: {ex}"
```
jpichon, line 253: Missing leading 'f' here too.

git_wrapper/branch.py
jpichon: No strong feelings either way! Probably why the discrepancy, heh. I like the change, thanks for the update.

git_wrapper/branch.py
jguiditta: Updated, thanks for catching those!  Hopefully that is all of them now.  While I was making this change, I realized we also have instances of the old '%s' style, but I am going to do a separate patch for that after this is in.

jpichon (approved):

fuzzball81 (approved):

PR #99, opened 2021-09-16
Replace %s string substitutions with f-strings.

### LINKED ISSUE #32: Drop Python 2 compatibility library

jguiditta:
A follow-up to the previous patch, should close #32.  If other
python 2.x syntax is found, we can remove as it is noticed.


fuzzball81 (approved):
Nice readability improvements

jpichon (approved):

PR #100, opened 2021-09-27
commit_format should be an internal-only method

jpichon:
commit_format takes in commits as represented by the underlying GitPython library, which should be abstracted away. End users should either call log_diff, short_log_diff or log_show_commit.


jguiditta (dismissed):
LGTM

jpichon:
(Rebased)

jguiditta (approved):

fuzzball81 (approved):

PR #102, opened 2021-10-13
Initial automation of changelog generation.

jguiditta:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,7 +28,8 @@ author = 'Jason Joyce, Julie Pichon'
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ['sphinx.ext.autodoc', 'sphinx.ext.viewcode', 'sphinx.ext.napoleon']
+extensions = ['sphinx.ext.autodoc', 'sphinx.ext.viewcode',
+              'sphinx.ext.napoleon', 'myst_parser']
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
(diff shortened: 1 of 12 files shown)


docs/conf.py
```diff
@@ -50,3 +56,19 @@
...
+import os
+import subprocess
+
+def setup(app):
+    app.connect('builder-inited', build_changelog)
+
+def build_changelog(app):
+    docs_dir = os.getcwd()
+    os.chdir('../')
+    cmd = ['bash', os.getcwd() + '/tooling/build_changelog', os.getcwd()]
```
fuzzball81, line 70: I would use an f-string for the middle argument.
It may also be worth pulling the `os.getcwd()` result to a variable so it's easy to reuse it since you make the call more than once.

tooling/build_changelog
```diff
@@ -0,0 +1,8 @@
+the_dir=$1
+: ${the_dir:=.}
+cd $the_dir
+pwd
+curl -s "https://raw.githubusercontent.com/release-depot/change/0.14.3/change" |
+sh -s -- init
+curl -s "https://raw.githubusercontent.com/release-depot/change/0.14.3/change" |
+sh -s --
```
fuzzball81, line 8: This looks like a repeat, am sure there is a reason but a comment or something explaining it would be helpful.

fuzzball81 (commented):
A few minor comments, otherwise it looks good.

tooling/build_changelog
jguiditta: Ok, I can add a comment on why this has to run twice, no problem.

fuzzball81 (approved):

jpichon (commented):
LGTM! Result looks good. I think we should also remove HISTORY.rst because it will rot, or maybe this is planned in another patch since the release steps etc will also need to be updated then?

PR #103, opened 2021-10-14
Update documentation to support new 'autochangelog' setup.

jguiditta:
* Removed unused HISTORY.rst
* Updated stale AUTHORS.rst


jpichon (commented):
Sorry about the integration tests shenanigans. I think we also need to remove HISTORY.rst from setup.cfg (and replace it with the changelog if that's possible, not sure because of the mix of rst and md. Maybe just a link in the README to the RTD changelog instead?). I'm surprised the twine check doesn't catch this, since it was meant to help up catch README/HISTORY formatting issues before uploading to pypi.

jpichon (approved):

PR #107, opened 2022-06-21
Added Prune Functionality To Fetch

siddhantdubey:
Prune related tags were added to the fetch
functions in order to give more functionality
to consumers. This will allow for branches and tags
to be easily pruned simply by calling the fetch function
with the appropriate boolean tags.

### DIFF
--- a/git_wrapper/remote.py
+++ b/git_wrapper/remote.py
@@ -52,10 +52,18 @@ class GitRemote(object):
 
         return ret_status
 
-    def fetch(self, remote="origin"):
-        """Refresh the specified remote
+    def fetch(self, remote="origin", prune=False, prune_tags=False):
+        """Refresh the specified remote.
+
+           Optionally, specify prune to be True to allow
+           for the removal of local branches that don't
+           exist on the remote. If you also wish to remove
+           local tags that don't exist on the remote, specify
+           prune_tags to be True.
 
            :param str remote: Remote to fetch
+           :param bool prune: True if you want to prune
+           :param bool prune_tags: True if you want to prune local tags (only works if prune is True)
         """
         try:
             remote = self.git_repo.repo.remote(remote)
@@ -65,24 +73,37 @@ class GitRemote(object):
             raise exceptions.ReferenceNotFoundException(msg)
 
         try:
-            remote.fetch()
+            if prune_tags and not prune:
+                self.logger.info("prune_tags was ignored because prune is False")
+                remote.fetch()
+            else:
+                remote.fetch(prune=prune, prune_tags=prune_tags)
         except git.GitCommandError as ex:
             msg = (f"Could not fetch remote {remote.name} ({remote.url}). "
                    f"Error: {ex}")
             raise exceptions.RemoteException(msg) from ex
 
-    def fetch_all(self):
+    def fetch_all(self, prune=False, prune_tags=False):
         """Refresh all the repo's remotes.
 
            All the remotes will be fetched even if one fails ; in this case a
            single exception containing the list of failed remotes is returned.
+
+           Optionally, specify prune to be True to allow
+           for the removal of local branches that don't
+           exist on the remotes. If you also wish to remove
+           local tags that don't exist on the remotes, specify
+           prune_tags to be True.
+
+           :param bool prune: True if you want to prune
+           :param bool prune_tags: True if you want to prune local tags (only works if prune is True)
         """
         remotes = self.names()
 
         errors = []
         for remote in remotes:
             try:
-                self.fetch(remote)
+                self.fetch(remote, prune, prune_tags)
             except exceptions.RemoteException:
                 self.logger.exception(f"Error fetching remote {remote}")
                 errors.append(remote)
--- a/integration_tests/test_remote.py
+++ b/integration_tests/test_remote.py
@@ -1,0 +1,90 @@
+from git_wrapper.repo import GitRepo
+
+
+def test_prune_tags(repo_root, clone_repo_root):
+    repo = GitRepo(repo_root)
+    cloned_repo = GitRepo.clone(f"file://{repo_root}", clone_repo_root)
+
+    cloned_repo.tag.create("test_tag", "master")
+    assert "test_tag" in cloned_repo.tag.names()
+
+    cloned_repo.tag.push("test_tag", "origin")
+    assert "test_tag" in repo.tag.names()
+
+    cloned_repo.git.push("origin", "--delete", "test_tag")
+    assert "test_tag" not in repo.tag.names()
+    assert "test_tag" in cloned_repo.tag.names()
+    cloned_repo.remote.fetch("origin", prune=True, prune_tags=True)
+    assert "test_tag" not in cloned_repo.tag.names()
+
+
+def test_no_prune(repo_root, clone_repo_root):
+    repo = GitRepo(repo_root)
+    cloned_repo = GitRepo.clone(f"file://{repo_root}", clone_repo_root)
+
+    cloned_repo.tag.create("test_tag", "master")
+    assert "test_tag" in cloned_repo.tag.names()
+
+    cloned_repo.tag.push("test_tag", "origin")
+    assert "test_tag" in repo.tag.names()
+
+    repo.tag.delete("test_tag")
+    assert "test_tag" not in repo.tag.names()
+
+    cloned_repo.remote.fetch("origin")
+    assert "test_tag" in cloned_repo.tag.names()
+
+
+def test_prune_branches(repo_root, clone_repo_root):
+    repo = GitRepo(repo_root)
+    cloned_repo = GitRepo.clone(f"file://{repo_root}", clone_repo_root)
+
+    cloned_repo.branch.create("test_branch", "master")
+    cloned_repo.tag.create("testing_tag", "master")
+    assert "test_branch" in [b.name for b in cloned_repo.repo.branches]
+    assert "testing_tag" in cloned_repo.tag.names()
+
+    # throughout the test, we'll make sure we never delete test_tag in the cloned repo
+
+    cloned_repo.git.push("origin", "test_branch")
+    cloned_repo.git.push("origin", "testing_tag")
+    assert "test_branch" in [b.name for b in repo.repo.branches]
+    assert "origin/test_branch" in cloned_repo.repo.references
+
+    repo.git.branch("-D", "test_branch")
+    repo.tag.delete("testing_tag")
+    assert "test_branch" not in [b.name for b in repo.repo.branches]
+    assert "testing_tag" not in repo.tag.names()
+    assert "origin/test_branch" in cloned_repo.repo.references
+
+    assert "testing_tag" in cloned_repo.tag.names()
+    cloned_repo.remote.fetch("origin", prune=True)
+    assert "origin/test_branch" not in cloned_repo.repo.references
+    assert "testing_tag" in cloned_repo.tag.names()
+
+
+def test_prune_both(repo_root, clone_repo_root):
+    repo = GitRepo(repo_root)
+    cloned_repo = GitRepo.clone(f"file://{repo_root}", clone_repo_root)
+
+    cloned_repo.branch.create("test_branch", "master")
+    cloned_repo.tag.create("testing_tag", "master")
+    assert "test_branch" in [b.name for b in cloned_repo.repo.branches]
+    assert "testing_tag" in cloned_repo.tag.names()
+
+    cloned_repo.git.push("origin", "test_branch")
+    cloned_repo.git.push("origin", "testing_tag")
+    assert "test_branch" in [b.name for b in repo.repo.branches]
+    assert "testing_tag" in repo.tag.names()
+    assert "origin/test_branch" in cloned_repo.repo.references
+
+    repo.git.branch("-D", "test_branch")
+    repo.tag.delete("testing_tag")
+    assert "test_branch" not in [b.name for b in repo.repo.branches]
+    assert "testing_tag" not in repo.tag.names()
+    assert "origin/test_branch" in cloned_repo.repo.references
+    assert "testing_tag" in cloned_repo.tag.names()
+
+    cloned_repo.remote.fetch("origin", prune=True, prune_tags=True)
+    assert "origin/test_branch" not in cloned_repo.repo.references
+    assert "testing_tag" not in cloned_repo.tag.names()
--- a/tests/test_remote.py
+++ b/tests/test_remote.py
@@ -101,8 +101,67 @@ def test_fetch(mock_repo):
 
     repo = GitRepo(repo=mock_repo)
     repo.remote.fetch()
-    assert mock_repo.remote.called is True
-    assert mock_remote.fetch.called is True
+
+    mock_repo.remote.assert_called()
+    mock_remote.fetch.assert_called()
+    mock_remote.fetch.assert_called_with(prune=False, prune_tags=False)
+
+
+def test_fetch_prune(mock_repo):
+    """
+    GIVEN GitRepo is initialized with a path and repo
+    WHEN remote.fetch is called
+    AND prune is True
+    THEN repo.remote is called
+    AND fetch is called with prune
+    """
+    mock_remote = Mock()
+    mock_repo.remote.return_value = mock_remote
+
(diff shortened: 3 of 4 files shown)


git_wrapper/remote.py
```diff
@@ -65,13 +66,13 @@ def fetch(self, remote="origin"):
...
         try:
-            remote.fetch()
+            remote.fetch(prune=prune, tags=tags, branches=branches)
         except git.GitCommandError as ex:
             msg = (f"Could not fetch remote {remote.name} ({remote.url}). "
                    f"Error: {ex}")
             raise exceptions.RemoteException(msg) from ex
 
-    def fetch_all(self):
+    def fetch_all(self, prune=False, tags=False, branches=False):
```
jpichon, line 75: The doc string will need to be updated to explain what the new arguments are and how to use them

git_wrapper/remote.py
```diff
@@ -52,7 +53,7 @@ def add(self, name, url):
 
         return ret_status
 
-    def fetch(self, remote="origin"):
+    def fetch(self, remote="origin", prune=False, tags=False, branches=False):
```
fuzzball81, line 56: Docstring needs to be updated to match the new arguments.

tests/test_remote.py
```diff
@@ -105,6 +105,42 @@ def test_fetch(mock_repo):
...
+    mock_remote = Mock()
+    mock_repo.remote.return_value = mock_remote
+
+    repo = GitRepo(repo=mock_repo)
+    repo.remote.fetch(prune=True, branches=True)
+    assert mock_repo.remote.called is True
+    assert mock_remote.fetch.called is True
+    assert mock_remote.fetch.call_args[1]['prune'] is True
+    assert mock_remote.fetch.call_args[1]['branches'] is True
+
```
fuzzball81, line 142: I would consider adding a test where prune, tags and branches are both true.

tests/test_remote.py
```diff
@@ -155,6 +191,54 @@ def test_fetch_all(mock_repo):
...
+    repo.remote.names = Mock(return_value=["origin", "otherremote"])
+
+    repo.remote.fetch_all(prune=True, branches=True)
+    assert mock_remoteA.fetch.called is True
+    assert mock_remoteB.fetch.called is True
+    assert mock_remoteA.fetch.call_args[1]['prune'] is True
+    assert mock_remoteA.fetch.call_args[1]['branches'] is True
+    assert mock_remoteB.fetch.call_args[1]['prune'] is True
+    assert mock_remoteB.fetch.call_args[1]['branches'] is True
+
```
fuzzball81, line 240: I would consider adding a test where prune, tags and branches are both true.

git_wrapper/remote.py
```diff
@@ -65,13 +66,13 @@ def fetch(self, remote="origin"):
             raise exceptions.ReferenceNotFoundException(msg)
 
         try:
-            remote.fetch()
+            remote.fetch(prune=prune, tags=tags, branches=branches)
```
jpichon, line 69: By this point, remote is a GitPython object, isn't it? 

```
(Pdb) type(remote)
<class 'git.remote.Remote'>
```

I don't think this can work, as the remote object's function wouldn't understand [any of these parameters](https://github.com/gitpython-developers/GitPython/blob/main/git/remote.py#L918)? We'd need to switch the arguments to the flags as understood by git-fetch, I believe.

Ideally to prove this works well, an [integration test](https://github.com/release-depot/git_wrapper/tree/master/integration_tests) would be created. Because a remote needs to be modified, this isn't quite straightforward but the trick used when [testing repo cloning](https://github.com/release-depot/git_wrapper/blob/master/integration_tests/conftest.py#L37) might help if we want to investigate this approach.

git_wrapper/remote.py
jpichon: In case you haven't come across it yet, we publish the docs and API references automatically. So the [method documentation for fetch](https://git-wrapper.readthedocs.io/en/latest/git_wrapper.html#git_wrapper.remote.GitRemote.fetch) will show the new arguments once they're added to the doc string.

jpichon (changes requested):
Sorry for the short review yesterday, I didn't have time for an in-depth one but was hoping an update with the doc string would help to inform the review as the parameter names and usage isn't straightforward. I made a few comments inline, I hope this makes sense.

git_wrapper/remote.py
```diff
@@ -32,7 +32,8 @@ def add(self, name, url):
             :return bool: True if the remote was added, False otherwise
         """
         working_dir = self.git_repo.repo.working_dir
-        self.logger.debug(f"Adding remote {name} ({url}) to repo {working_dir}")
+        self.logger.debug(
+            f"Adding remote {name} ({url}) to repo {working_dir}")
```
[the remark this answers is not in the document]
siddhantdubey: Sorry, my editor's auto formatting on save messed this up. I'll change my local settings.

git_wrapper/remote.py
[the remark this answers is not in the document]
siddhantdubey: You are correct, only two boolean parameters are required. I think a better name for the tags variable would then be prune_tags. I'll change the function and the docstring to reflect this.

git_wrapper/remote.py
siddhantdubey: Remote is a GitPython object, and the parameters I have right now are incorrect but the correct parameters look like this:
`remote.fetch(prune=prune,prune_tags=tags)`
Which leads to the following git command being executed when it's used in `git_tag_cleaner`.
`INFO:git.cmd:git fetch -v --prune --prune-tags origin`
GitPython isn't super thoroughly documented, but I found out that they allow that by searching around on the internet.

jpichon: Looks good!

bsarabeswaran:
LGTM!

git_wrapper/remote.py
```diff
@@ -65,24 +67,27 @@ def fetch(self, remote="origin"):
...
-    def fetch_all(self):
+    def fetch_all(self, prune=False, prune_tags=False):
         """Refresh all the repo's remotes.
 
            All the remotes will be fetched even if one fails ; in this case a
            single exception containing the list of failed remotes is returned.
+
+           :param bool prune: True if you want to prune
+           :param bool prune_tags: True if you want to prune tags
         """
```
jpichon, line 84: Similar explanation here. Maybe it can be inspired from the man page explanation for the --prune parameter, so library consumers really understand what they're doing when they're deciding whether to use the option.

tests/test_remote.py
```diff
@@ -105,6 +105,43 @@ def test_fetch(mock_repo):
     assert mock_remote.fetch.called is True
 
 
+def test_fetch_prune(mock_repo):
+    """
+    GIVEN GitRep is initialized with a path and repo
```
jpichon, line 110: GitRep -> GitRepo

tests/test_remote.py
```diff
@@ -105,6 +105,43 @@ def test_fetch(mock_repo):
...
+    """
+    mock_remote = Mock()
+    mock_repo.remote.return_value = mock_remote
+
+    repo = GitRepo(repo=mock_repo)
+    repo.remote.fetch(prune=True)
+    assert mock_repo.remote.called is True
+    assert mock_remote.fetch.called is True
+    assert mock_remote.fetch.call_args[1]['prune'] is True
+
```
jpichon, line 124: We probably want to check that prune_tags is not in the list of arguments as well, to make sure it doesn't get sent by mistake.

jpichon (commented):
I made a few small comments and suggestions inline, but I haven't tested this manually yet. I know you mentioned you were working on an integration test, I think it would be great to have it as part of this patch to increase the confidence in its correctness now and in the future. Thank you.

git_wrapper/remote.py
jpichon: I suppose "remote" should be plural in the explanation here, but this is good otherwise!

integration_tests/test_remote.py
```diff
@@ -0,0 +1,65 @@
+from git_wrapper.repo import GitRepo
+
+
+def test_prune_tags(repo_root):
```
jpichon, line 4: I don't have the time to do a full review yet so just a couple of quick notes on this one test! I love to see the work around the integration tests by the way, having this will really help.

integration_tests/test_remote.py
```diff
@@ -0,0 +1,65 @@
...
+
+    repo.tag.create("test_tag", "master")
+    assert "test_tag" in repo.tag.names()
+
+    repo.tag.push("test_tag", "test")
+
+    repo.git.push("test", "--delete", "test_tag")
+
+    repo.remote.fetch("test", prune=True, prune_tags=True)
+    assert "test_tag" not in repo.tag.names()
```
jpichon, line 19: Unfortunately, this doesn't quite work as is: if you remove L18, the test still pass.

I think you will need 2 repositories for this one, repo_root can be one and a second one that clones from it using the file system approach you have on L7 (the [clone tests](https://github.com/release-depot/git_wrapper/blob/master/integration_tests/test_repo.py#L33) might provide inspiration too, although it's okay to use GitPython directly for this particular test here). Then you can make changes to the main repo_root repo directly (no need to push, just delete at will) and see if cloned_repo.fetch() prunes what it should after that.

integration_tests/test_remote.py
```diff
@@ -0,0 +1,65 @@
...
+
+def test_prune_tags(repo_root):
+    repo = GitRepo(repo_root)
+
+    repo.remote.add("test", f"file://{repo_root}")
+    remotes = repo.remote.names()
+    assert "test" in remotes
+
+    repo.tag.create("test_tag", "master")
+    assert "test_tag" in repo.tag.names()
```
jpichon, line 12: This is great btw, checking that we have the data we need where it should first before acting on it is a great starting point.

jpichon (changes requested):
A quick partial review for now, with hopefully some helpful advice!

When you have a chance, if you'd like to open a separate PR to update the [CONTRIBUTING](https://github.com/release-depot/git_wrapper/blob/master/CONTRIBUTING.rst) doc with some information on how to run the integration tests locally with podman, so that the next person who joins the project doesn't hit the same issues you did, that would be very appreciated. Thank you!

git_wrapper/remote.py
```diff
@@ -52,10 +52,12 @@ def add(self, name, url):
 
         return ret_status
 
-    def fetch(self, remote="origin"):
+    def fetch(self, remote="origin", prune=False, prune_tags=False):
         """Refresh the specified remote
 
            :param str remote: Remote to fetch
+           :param bool prune: True if you want to prune
+           :param bool prune_tags: True if you want to prune tags
```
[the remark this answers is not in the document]
jpichon: Any thoughts about this part?

> then in the code, ignore the flag and don't pass it to remote.fetch() if prune_tags is provided on its own without prune. (Just log that we're ignoring it in that case.)

I think this could be 3 lines or so, if prune is False but prune_tags is True, set prune_tags to False and log at info level that prune-tags will be ignored. I'm wary about the "strange interactions" described in the man-page when it's used on its own, as the command might expand again in the future.

jpichon (commented):
Thank you for the updates! This is looking in good shape.

integration_tests/test_remote.py
```diff
@@ -0,0 +1,16 @@
...
+    repo = GitRepo(repo_root)
+    cloned_repo = GitRepo.clone(f"file://{repo_root}", clone_repo_root)
+
+    cloned_repo.tag.create("test_tag", "master")
+    assert "test_tag" in cloned_repo.tag.names()
+
+    cloned_repo.tag.push("test_tag", "origin")
+    assert "test_tag" in repo.tag.names()
+
+    cloned_repo.git.push("origin", "--delete", "test_tag")
```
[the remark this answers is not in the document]
siddhantdubey: I'll add the branch pruning tests back.

git_wrapper/remote.py
```diff
@@ -65,24 +73,38 @@ def fetch(self, remote="origin"):
...
         """Refresh all the repo's remotes.
 
            All the remotes will be fetched even if one fails ; in this case a
            single exception containing the list of failed remotes is returned.
+
+           Optionally, specify prune to be True to allow
+           for the removal of local branches that don't
+           exist on the remotes. If you also wish to remove
+           local tags that don't exist on the remotes, specify
+           prune_tags to be true.
```
eggmaster, line 97: <nit> True is capitalized elsewhere

eggmaster (commented):
This looks pretty good to me and I can +1 after jpichon completes her review.

integration_tests/test_remote.py
```diff
@@ -0,0 +1,92 @@
...
+    repo = GitRepo(repo_root)
+    cloned_repo = GitRepo.clone(f"file://{repo_root}", clone_repo_root)
+
+    cloned_repo.tag.create("test_tag", "master")
+    assert "test_tag" in cloned_repo.tag.names()
+
+    cloned_repo.tag.push("test_tag", "origin")
+    assert "test_tag" in repo.tag.names()
+
+    cloned_repo.remote.fetch("origin")
```
jpichon, line 31: I think we need to delete on the remote before fetching, to prove that fetch didn't delete anything.

integration_tests/test_remote.py
```diff
@@ -0,0 +1,92 @@
...
+    cloned_repo.tag.create("testing_tag", "master")
+    assert "test_branch" in [b.name for b in cloned_repo.repo.branches]
+    assert "testing_tag" in cloned_repo.tag.names()
+
+    # throughout the test, we'll make sure we never delete test_tag in the cloned repo
+
+    cloned_repo.git.push("origin", "test_branch")
+    cloned_repo.git.push("origin", "testing_tag")
+    assert "test_branch" in [b.name for b in repo.repo.branches]
+    assert "origin/test_branch" in [r.name for r in cloned_repo.repo.remotes.origin.refs]
```
jpichon, line 56: (Nit) If you wanted to simplify you could simply check in `cloned_repo.repo.references` but not a blocker if you prefer this way.

integration_tests/test_remote.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def test_prune_both(repo_root, clone_repo_root):
+    repo = GitRepo(repo_root)
+    cloned_repo = GitRepo.clone(f"file://{repo_root}", clone_repo_root)
+
+    cloned_repo.branch.create("test_branch", "master")
+    cloned_repo.tag.create("test", "master")
+    assert "test_branch" in [b.name for b in cloned_repo.repo.branches]
+    assert "test" in cloned_repo.tag.names()
```
jpichon, line 76: This is also a nit, feel free to ignore. Personally, I think mimicking the pattern of the previous test, in which the testing tag is called `testing_tag` makes it easier to spot the difference between the two tests, and also makes it easier on the reader to remember which is a branch and which is a tag.

integration_tests/test_remote.py
```diff
@@ -0,0 +1,92 @@
...
+    cloned_repo.git.push("origin", "test")
+    assert "test_branch" in [b.name for b in repo.repo.branches]
+    assert "test" in repo.tag.names()
+    assert "origin/test_branch" in [r.name for r in cloned_repo.repo.remotes.origin.refs]
+
+    repo.git.branch("-D", "test_branch")
+    repo.tag.delete("test")
+    assert "test_branch" not in [b.name for b in repo.repo.branches]
+    assert "test" not in repo.tag.names()
+    assert "origin/test_branch" in [r.name for r in cloned_repo.repo.remotes.origin.refs]
```
jpichon, line 88: In addition to the branch, you could also confirm the tag is in the cloned_repo before fetching, here and in the previous test, if you like.

jpichon (changes requested):
I really like this, thank you for the updates! I noted a few nits inline you're free to ignore if you prefer. I'd just like to see a unit test for the new "ignore prune_tags if prune=False" path (no need for integration, a quick running unit test is enough) and this looks ready to me after that.

jpichon (dismissed):
LGTM! Someone with fresher eyes can do the second review :)

git_wrapper/remote.py
eggmaster: Can you make the s/true/True/ change here? Consistency is good. Otherwise I am ready to approve, nice work!

eggmaster (approved):

PR #108, opened 2022-06-27
Added Information About Integration Tests To Docs

siddhantdubey:
CONTRIBUTING.rst was missing information on how to run integration
tests using podman, so I added information about the command that
runs the integration tests.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -77,10 +77,11 @@ Ready to contribute? Here's how to set up `git_wrapper` for local development.
    Now you can make your changes locally.
 
 5. When you're done making changes, check that your changes pass flake8 and the
-   tests, including testing other Python versions with tox::
+   tests, including testing other Python versions with tox, and running integration tests::
 
     $ make lint
     $ make test
+    $ tox -e integration_podman
     $ tox
 
 


CONTRIBUTING.rst
```diff
@@ -77,10 +77,11 @@ Ready to contribute? Here's how to set up `git_wrapper` for local development.
    Now you can make your changes locally.
 
 5. When you're done making changes, check that your changes pass flake8 and the
-   tests, including testing other Python versions with tox::
+   tests, including testing other Python versions with tox, and running integration tests::
 
     $ make lint
     $ make test
+    $ make -eintegration_podman
```
eggmaster, line 84: I think there should be a space after '-e', but even with that I get:
make: *** No rule to make target 'integration_podman'.  Stop.

siddhantdubey: You're right the command should be tox -eintegration_podman, which needs to be run in the pipenv. I'll update it to show that.

CONTRIBUTING.rst
```diff
@@ -77,10 +77,11 @@ Ready to contribute? Here's how to set up `git_wrapper` for local development.
    Now you can make your changes locally.
 
 5. When you're done making changes, check that your changes pass flake8 and the
-   tests, including testing other Python versions with tox::
+   tests, including testing other Python versions with tox, and running integration tests::
 
     $ make lint
     $ make test
+    $ tox -eintegration_podman
```
fuzzball81, line 84: This is more of a personal preference. I would add a space between -e and 'integration_podman' it just makes it easier to read.

fuzzball81 (approved):

PR #113, opened 2025-02-12
Fix the format of the readthedocs config file.

jguiditta:
Docs were no longer building due to a schema change on readthedocs.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -5,12 +5,15 @@
 # # Required
 version: 2
 
+build:
+  os: ubuntu-24.04
+  tools:
+    python: "3.12"
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
    configuration: docs/conf.py
 
 python:
-    version: "3.8"
     install:
       - requirements: docs-requirements.txt
       - requirements: requirements.txt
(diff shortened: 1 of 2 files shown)


jguiditta:
Changes build docs successfully. They can be seen here:
https://test-git-wrapper.readthedocs.io/en/latest/

jguiditta:
> LGTM ;)
> 
> I'm not seeing the "Changelog" link in the menu of the test docs vs [the current docs](https://git-wrapper.readthedocs.io/en/latest/changelog.html), when I think it's quite handy to have, but not sure if it's just a test vs real env difference or if there was another format change somewhere.

This is a good observation, and thanks for the review! I did some looking, and I believe this is just because of how I have my test project set up on readthedocs. I have it pointed to a specific branch, and it does a shallow clone, so there are no tags. This causes our tooling which generates the changelog to give this output:

````
change stderr returns: couldn't find any valid version tags
couldn't find CHANGELOG.md
```

I was able to verify that no tags are present in the checkout when done the way test rtd does it. If we see this is still misbehaving on merge, I will follow up with another patch fixing it, if no objections.

jpichon:
That sounds absolutely reasonable, thank you for checking and for the explanation!

evallesp (approved):
LGTM

eggmaster (approved):
LGTM, thanks for fixing this!

PR #114, opened 2025-03-14
Remove remote for re-creating it if url has changed

evallesp:
We're adding the possibility to remove the remote. This, with the addition of names_url_dict, makes the posibility of checking the remote url, and if it's updated, then remove it for re-creating.

### DIFF
--- a/git_wrapper/remote.py
+++ b/git_wrapper/remote.py
@@ -24,6 +24,14 @@ class GitRemote(object):
         """
         return [x.name for x in self.git_repo.repo.remotes]
 
+    def names_url_dict(self):
+        """Returns a dict of remotes for a given repo with its url
+
+            :return dict: A dict of utf-8 encoded remote names with
+                          url as its value
+        """
+        return {x.name: x.url for x in self.git_repo.repo.remotes}
+
     def add(self, name, url):
         """Adds a remote to the given repo
 
@@ -111,3 +119,28 @@ class GitRemote(object):
         if errors:
             msg = f"Error fetching these remotes: {', '.join(errors)}"
             raise exceptions.RemoteException(msg)
+
+    def remove(self, name):
+        """Remove the specified remote from the given repo
+
+           :param str name: Remote name to remove
+           :return bool: True if the remote was removed, False otherwise
+        """
+        working_dir = self.git_repo.repo.working_dir
+        self.logger.debug(f"Removing remote {name} from repo {working_dir}")
+        ret_status = False
+
+        try:
+            remote = self.git_repo.repo.remote(name)
+        except ValueError:
+            repo = self.git_repo.repo.working_dir
+            msg = f"Remote {name} does not exist on repo {repo}"
+            raise exceptions.ReferenceNotFoundException(msg)
+
+        try:
+            self.git_repo.repo.delete_remote(remote)
+            ret_status = True
+        except git.CommandError:
+            return ret_status
+
+        return ret_status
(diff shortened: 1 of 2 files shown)


tests/test_remote.py
```diff
@@ -89,6 +115,47 @@ def test_add_remote_update_fails(mock_repo):
     delete_mock.assert_called_once_with(remote_mock)
 
 
+def test_remove_remote_removes(mock_repo):
+    """
+    GIVEN GitRepo initialized with a path and repo
+    WHEN remote.add is called with a name and url
```
jguiditta, line 121: question: Shouldn't this (and the following ones) be 'remote.remove'?

jguiditta (commented):
Couple minor things, otherwise seems reasonable

tests/test_remote.py
```diff
@@ -21,6 +21,17 @@ def remote_generator(names):
     return ret_data
 
 
+def remote_generator_url(remotes):
+    """Generates objects to be used with git.Repo.remotes call"""
+    ret_data = []
+    for name, url in remotes.items():
+        obj = type('', (), {})()
```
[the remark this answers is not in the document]
eggmaster: I agree with jguiditta here, and I think declaring a dataclass in this method is probably less work than writing a comment and creating a follow-on task! But a comment would suffice for the moment.

eggmaster (commented):
couple comments in addition/response to jguiditta's, but this looks good to me otherwise!

git_wrapper/remote.py
```diff
@@ -111,3 +119,28 @@ def fetch_all(self, prune=False, prune_tags=False):
...
             raise exceptions.RemoteException(msg)
+
+    def remove(self, name):
+        """Remove the specified remote to the given repo
+
+           :param str name: Remote name to remove
+           :return bool: True if the remote was removed, False otherwise
+        """
+        working_dir = self.git_repo.repo.working_dir
+        self.logger.debug(f"Removing remote {name} to repo {working_dir}")
```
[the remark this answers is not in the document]
evallesp: Thanks!

tests/test_remote.py
```diff
@@ -10,14 +11,26 @@
...
+    url: str
+
 def remote_generator(names):
     """Generates objects to be used with git.Repo.remotes call"""
     ret_data = []
     for name in names:
-        obj = type('', (), {})()
-        obj.name = name
-        ret_data.append(obj)
+        ret_data.append(RemoteNames(name))
```
evallesp, line 26: let me know what you think about this. I found that this refactor might suit here.

eggmaster (dismissed):

tests/test_remote.py
jguiditta: I like this much better, thanks for making this change!

tests/test_remote.py
```diff
@@ -89,6 +120,47 @@ def test_add_remote_update_fails(mock_repo):
     delete_mock.assert_called_once_with(remote_mock)
 
 
+def test_remove_remote_removes(mock_repo):
+    """
+    GIVEN GitRepo initialized with a path and repo
+    WHEN remote.remove is called with a name and url
+    THEN a TRUE status is returned
+    WITH update called
```
jguiditta, line 128: question: I don't see any assertion related to this?

jguiditta (commented):
One question, otherwise, I think this is good!

tests/test_remote.py
evallesp: You're right. Removed and updated docstring after revisit.

jguiditta (approved):

eggmaster (approved):

